PR #82, opened 2023-03-26
Ignore the invalid audio frames

### LINKED ISSUE #81: [mp3float] Header missing
Hi first of all thank you for this great repo!

## Issue
Transcribing some mp3 files I got the following error: " [mp3float] Header missing"

## The cause:  
"faster_whisper/audio.py", line 42, in decode_audio", I saw faster whisper uses PyAv instead of the FFmpeg package whisper uses, I'm a total noob when it comes to FFmpeg, but I'd be more than happy to help if I can.

## How I got it working:

I'm using the following Load audio from whisper, to first take the mp3 file as input:
```python
def load_audio(file: str, sr: int = SAMPLE_RATE):
    """
    Open an audio file and read as mono waveform, resampling as necessary
    Parameters
    ----------
    file: str
        The audio file to open
    sr: int
        The sample rate to resample the audio if necessary
    Returns
    -------
    A NumPy array containing the audio waveform, in float32 dtype.
    """
    try:
        # This launches a subprocess to decode audio while down-mixing and resampling as necessary.
        # Requires the ffmpeg CLI and `ffmpeg-python` package to be installed.
        out, _ = (
            ffmpeg.input(file, threads=0)
            .output("-", format="s16le", acodec="pcm_s16le", ac=1, ar=sr)
            .run(cmd=["ffmpeg", "-nostdin"], capture_stdout=True, capture_stderr=True)
        )
    except ffmpeg.Error as e:
        raise RuntimeError(f"Failed to load audio: {e.stderr.decode()}") from e

    return np.frombuffer(out, np.int16).flatten().astype(np.float32) / 32768.0
```

I then pass the returned np array to faster-whisper:
```python
audio = load_audio(file)
segments, info = model.transcribe(audio, **options_dict)
```
## Fix?
Doing this solves my problem, but I'd like to know if there are tweaks we can do with PyAv instead, so I don't have to use two different packages.

guillaumekln:
Fixes #81.

### DIFF
--- a/faster_whisper/audio.py
+++ b/faster_whisper/audio.py
@@ -36,6 +36,7 @@ def decode_audio(input_file: Union[str, BinaryIO], sampling_rate: int = 16000):
 
     with av.open(input_file, metadata_errors="ignore") as container:
         frames = container.decode(audio=0)
+        frames = _ignore_invalid_frames(frames)
         frames = _group_frames(frames, 500000)
         frames = _resample_frames(frames, resampler)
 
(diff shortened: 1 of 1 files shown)


Hannes1:
Can confirm this fixes my problem, thanks for the quick update!

PR #103, opened 2023-04-02
Added retrieval mechanism (avg_log_prob/no_speech_prob)

palladium123:


palladium123:
Sorry not sure how to fix the Black error.

guillaumekln:
You should install and run the `black` tool on the code:

```
pip install black
black faster_whisper/
```

faster_whisper/transcribe.py
```diff
@@ -482,7 +486,7 @@ def generate_with_fallback(
                 max_initial_timestamp_index=max_initial_timestamp_index,
                 **kwargs,
             )[0]
-
+            no_speech_prob = result.no_speech_prob
```
guillaumekln, line 489: Since `no_speech_prob` is included in `result` which is already returned by the function, maybe you don't need to return `no_speech_prob` separately?

guillaumekln (approved):

PR #121, opened 2023-04-06
Review Git Files, Setup Helper Script, Dev Requirements (including pytest)

bekirbakar:
This pull request includes the following changes:

- Added LFS support for audio files. This will allow us to store large audio files in our repository without causing issues with performance or storage. We may add more samples in the future to test performance with longer files.
- Configured .gitignore to exclude cache, editor files, and other non-essential files from the repository.
- Added a bash script to create a virtual environment and install all required packages. There are multiple requirement files for this project, so this script will simplify the setup process for contributors.

These changes were made to improve the organization and accessibility of the repository.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*.wav filter=lfs diff=lfs merge=lfs -text
+*.flac filter=lfs diff=lfs merge=lfs -text
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,19 @@
+# Byte-compiled / Optimized / DLL Files
 *.pyc
+*.pyo
+*.pyd
+__pycache__/
+
+# Distribution / Packaging
+venv/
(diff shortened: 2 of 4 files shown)


guillaumekln (commented):
Thank you for contributing! However, I'm sorry to say that I don't find these changes useful:

* We don't plan to push large audio files so I don't want to enable LFS for this repository
* We already have an extra requirement `dev` which can be used like `pip install -e .[dev]`
* The environment setup script is too specific, each user has its own process

I could accept the `.gitignore` change if you remove the audio extensions. Feel free to send another PR if you want to send this change only.

bekirbakar:
Thanks, @guillaumekln for the detailed feedback. I agree.

I started working with your repository and those were the things that I had to do as a first step. I thought it would be useful for other developers.

PR #122, opened 2023-04-06
Configure git ignore for more files.

bekirbakar:
Changes on `.gitignore` to exclude cache, editor files, and other non-essential files from the repository.


guillaumekln:
Thanks!

PR #126, opened 2023-04-08
Expose download location in WhisperModel constructor

entn-at:
This increases compatibility with OpenAI Whisper's [`whisper.load_model()`](https://github.com/openai/whisper/blob/b80bcf610d89960bc658b61af9c333fc6d978d78/whisper/__init__.py#L97) and is useful for downstream integrations


guillaumekln:
Thanks!

PR #154, opened 2023-04-16
Align segment structure with openai/whisper

tekacs:
At the moment, the segments returned by faster-whisper are a subtly different shape to those returned by Whisper and also lack some of the information that JSON/verbose transcription results from that repo provide.

It appears that all of the information necessary to return these is in fact present in the transcribe function of faster-whisper, but they're simply not returned at present.

This PR adds those in and adjusts the naming of avg_log_prob to be consistent with Whisper.

The quirk I'm not certain of is whether it would be best to adjust the returned `seek` value on the other side of the VAD, but I don't do that at present.

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -504,7 +504,6 @@ class WhisperModel:
             encoder_output = None
 
             for segment in current_segments:
-                idx += 1
                 tokens = segment["tokens"]
                 text = tokenizer.decode(tokens)
 
@@ -512,6 +511,7 @@ class WhisperModel:
                     continue
 
                 all_tokens.extend(tokens)
+                idx += 1
 
                 yield Segment(
                     id=idx,
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -333,6 +333,7 @@ class WhisperModel:
         encoder_output: Optional[ctranslate2.StorageView] = None,
     ) -> Iterable[Segment]:
         content_frames = features.shape[-1] - self.feature_extractor.nb_max_frames
+        idx = 0
         seek = 0
         all_tokens = []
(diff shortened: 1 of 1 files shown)


palladium123:
I created a PR for this in the past (before VAD was implemented), but I only implemented avg_logprob (sorry @guillaumekln  typo there in my original PR) and no_speech_prob as I at that time I didn't think the other values will change things from a decision making perspective (whether to reject the output or not, or to assess the reliability of the output). In hindsight, I could see how knowing the temperature of the output can be useful (as temperature >0 may not be reliable). 

I do wonder if VAD actually makes no_speech_prob (kinda) obsolete.....

faster_whisper/transcribe.py
```diff
@@ -504,16 +512,21 @@ def generate_segments(
                 all_tokens.extend(tokens)
 
                 yield Segment(
+                    id=idx,
```
guillaumekln, line 515: The id goes back to 0 when moving to the next 30-second window. You should increment a global index instead.

faster_whisper/transcribe.py
```diff
@@ -574,7 +588,7 @@ def generate_with_fallback(
             # Recover the average log prob from the returned score.
             seq_len = len(tokens)
             cum_log_prob = result.scores[0] * (seq_len**options.length_penalty)
```
guillaumekln, line 590: Can you rename `cum_log_prob` to `cum_logprob` for consistency with `avg_logprob`?

guillaumekln (changes requested):
Thanks! It makes sense to match the openai/whisper format even though I think some attributes are not very useful (especially `id` which is trivial to compute).

Regarding `seek` I don't think it is necessary to adjust it based on VAD. It will just make things more complex.

guillaumekln (approved):

PR #163, opened 2023-04-19
Ignore repeated prompt

heimoshuiyu:
Same with <https://github.com/openai/whisper/pull/1253>

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -330,6 +330,7 @@ class WhisperModel:
         content_frames = features.shape[-1] - self.feature_extractor.nb_max_frames
         seek = 0
         all_tokens = []
+        all_prompt_text = []
         prompt_reset_since = 0
 
         if options.initial_prompt is not None:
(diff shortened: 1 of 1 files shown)


archive-r:
In my case, there was a lot of improvement.

archive-r:
Hi, your commit brings a lot of improvements, but still sometimes the model gets stuck in an iterative loop.

As a more aggressive alternative, what do you think about adjusting prompt_reset_since when duplicate text is detected, as shown below?

```
            duplicate_prompt = False

            for segment in current_segments:
                tokens = segment["tokens"]
                text = tokenizer.decode(tokens)

                if segment["start"] == segment["end"] or not text.strip():
                    continue

                check_prompt_num = 1
                if all(
                    [
                        text.strip() != i.strip()
                        for i in all_prompt_text[-check_prompt_num:]
                    ]
                ):
                    all_tokens.extend(tokens)
                    all_prompt_text.append(text)
                else:
                    duplicate_prompt = True

                idx += 1

                yield Segment(
                    id=idx,
                    seek=seek,
                    start=segment["start"],
                    end=segment["end"],
                    text=text,
                    tokens=tokens,
                    temperature=temperature,
                    avg_logprob=avg_logprob,
                    compression_ratio=compression_ratio,
                    no_speech_prob=result.no_speech_prob,
                    words=(
                        [Word(**word) for word in segment["words"]]
                        if options.word_timestamps
                        else None
                    ),
                )

            if (
                not options.condition_on_previous_text
                or temperature > 0.5
                or duplicate_prompt
            ):
                prompt_reset_since = len(all_tokens)
```

Purfview:
Works well, and the users reported positive experience with this PR.

PR #170, opened 2023-04-23
Add transcription_options to AudioInfo

FlippFuzz:
It would be great if we can include the transcription_options in AudioInfo.

My application is only making a few changes but leaving the rest as default. However, I would like to record down all settings (including those that I did not specify) so that the audio can be transcribed again identically in future if need be.


guillaumekln:
Thanks. I will probably rename `AudioInfo` to something else now. For example `TranscriptionInfo`.

PR #175, opened 2023-04-24
Fix where tokens are reset

### LINKED ISSUE #138: Issue with condition_on_previous_text param ?
Hello Guillaume,

Either I misunderstood the condition_on_previous_text option or there is a small error in the implementation. I thought with this option no previous decoded text would be used but it appears text from the previous segment is used.
https://github.com/guillaumekln/faster-whisper/blob/2b53dee6b697ba1364d7ba1b1c1ce3a6a2dc05df/faster_whisper/transcribe.py#L191-L194

If it is a mistake it is triggered by this line that should be at the end of the for loop / after extending all_tokens:
https://github.com/guillaumekln/faster-whisper/blob/2b53dee6b697ba1364d7ba1b1c1ce3a6a2dc05df/faster_whisper/transcribe.py#L470-L471

Best,
Valentin

mayeaux:
Fixes #138 

I believe this is correct, I wanted to test it so I implemented a PR for it

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -483,9 +483,6 @@ class WhisperModel:
 
                 seek += segment_size
 
-            if not options.condition_on_previous_text or temperature > 0.5:
-                prompt_reset_since = len(all_tokens)
-
             if options.word_timestamps:
                 self.add_word_timestamps(
                     current_segments,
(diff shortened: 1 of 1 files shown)


mayeaux:
Seems to work, btw. Makes a big difference.

Issue #194, opened 2023-04-28
Enhancement: Add VAD options into TranscriptionInfo


FlippFuzz:
We've added the whisper transcription options into TranscriptionInfo in https://github.com/guillaumekln/faster-whisper/pull/170 and https://github.com/guillaumekln/faster-whisper/pull/174.
Is it possible to also add the VAD options to the TranscriptionInfo?

I don't mind making a pull request, but it looks like we probably need to create new classes such as VadOptions, and possibly adjust the parameters for get_speech_timestamps() to use the new class.
I'm not 100% certain how you would like to go about doing this.

PR #197, opened 2023-04-29
Implementing #194

FlippFuzz:
1. I've created a new `VadOptions` class in vad.py to hold the list of options and default values.
2. I've also changed the parameters of `get_speech_timestamps` to use the VadOptions class.
I know that #194 said that perhaps we should leave the parameters as-is, but it really looks strange to have the `VadOptions` class defined at the top with defaults, and the `get_speech_timestamps` to not have any defaults.
3. `TranscriptionInfo` now also contains `VadOptions`.
4. Edited the `transcribe` and `get_speech_timestamps` documentation to mention the `VadOptions` class.

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -17,6 +17,7 @@ from faster_whisper.vad import (
     SpeechTimestampsMap,
     collect_chunks,
     get_speech_timestamps,
+    VadOptions
 )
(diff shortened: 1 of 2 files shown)


FlippFuzz:
Too many errors. 
Will create a new pull request when I have a working copy that is tested working locally.

PR #210, opened 2023-05-08
transcribe: return all language probabilities if requested

ozancaglayan:
This change adds a `all_language_probs` field to `TranscriptionInfo` so that when language detection is done, we can recover the full probabilities from the model in addition to the top one.

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -190,7 +190,7 @@ class WhisperModel:
           language: The language spoken in the audio. It should be a language code such
             as "en" or "fr". If not set, the language will be detected in the first 30 seconds
             of audio.
-          return_all_language_probs: If `True`, `TranscriptInfo` will return all language
+          return_all_language_probs: If `True`, `TranscriptionInfo` will return all language
             probabilities in addition to the top one.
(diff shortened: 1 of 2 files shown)


faster_whisper/transcribe.py
```diff
@@ -150,6 +151,7 @@ def transcribe(
         self,
         audio: Union[str, BinaryIO, np.ndarray],
         language: Optional[str] = None,
+        return_all_language_probs: bool = False,
```
guillaumekln, line 154: I'm OK to always return the language probabilities when they are available. So I suggest to simply remove this argument.

guillaumekln (changes requested):
I think the new argument is not needed.

faster_whisper/transcribe.py
ozancaglayan: Sure, thanks! To confirm, we'll still keep the top language name and prob for backward compatibility right?

guillaumekln: Yes, keep the existing properties and add `all_language_probs` (which can still be `None` if the language detection did not run).

ozancaglayan:
Test fails because Huggingface hub is down since some hours.

PR #215, opened 2023-05-09
rety model download locally if huggingface throws an http error.

daxaxelrod:
@guillaumekln

### DIFF
--- a/faster_whisper/utils.py
+++ b/faster_whisper/utils.py
@@ -77,8 +77,8 @@ def download_model(
         "tokenizer.json",
         "vocabulary.txt",
     ]
-    kwargs['allow_patterns'] = allow_patterns
-    kwargs['tqdm_class'] = disabled_tqdm
+    kwargs["allow_patterns"] = allow_patterns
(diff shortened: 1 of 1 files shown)


faster_whisper/utils.py
```diff
@@ -77,13 +77,20 @@ def download_model(
...
-        allow_patterns=allow_patterns,
-        tqdm_class=disabled_tqdm,
-        **kwargs,
-    )
+    try:
+        return huggingface_hub.snapshot_download(
+            repo_id,
+            **kwargs,
+        )
+    except huggingface_hub.utils._errors.HfHubHTTPError:
```
guillaumekln, line 88: You can use the public symbol here.

```suggestion
    except huggingface_hub.utils.HfHubHTTPError:
```

https://huggingface.co/docs/huggingface_hub/v0.14.1/en/package_reference/utilities#huggingface_hub.utils.HfHubHTTPError

guillaumekln (changes requested):

guillaumekln (approved):

PR #219, opened 2023-05-09
Correction for README.md

Purfview:
OpenAI is using beam_size 5 too.
I think you meant CPP/Const-me implementations.


guillaumekln:
No, the README is correct. See https://github.com/guillaumekln/faster-whisper/issues/6.

PR #284, opened 2023-06-07
Update README with community repo using FW

hedrergudene:
This contribution aims to bring faster-whisper users to explore community contributions that are fully scalable and production-ready. This is a piece of work that has already been implemented in several use cases and open sourcing the code means that anyone can quickly make the most of our joint efforts :)


guillaumekln:
Thanks!

PR #306, opened 2023-06-17
Allow users to input an Iterable of token ids into initial_prompt

FlippFuzz:
Discussion in https://github.com/guillaumekln/faster-whisper/issues/299#issuecomment-1594773275

This pull request allows users to specify an Iterable of token IDs for initial_prompt.

There is a minor deviation from that comment.
Instead of a list as an input, I've allowed `Iterable[int]`.

I think users would probably want to use a rotating array(such as `deque(maxlen=224)`) to automatically drop off old tokens.
Also, both lists and deques are iterable and can be input into the method.

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -362,12 +362,12 @@ class WhisperModel:
         prompt_reset_since = 0
 
         if options.initial_prompt is not None:
-            if isinstance(options.initial_prompt, Iterable):
-                all_tokens.extend(options.initial_prompt)
-            else:
+            if isinstance(options.initial_prompt, str):
                 initial_prompt = " " + options.initial_prompt.strip()
                 initial_prompt_tokens = tokenizer.encode(initial_prompt)
                 all_tokens.extend(initial_prompt_tokens)
+            else:
+                all_tokens.extend(options.initial_prompt)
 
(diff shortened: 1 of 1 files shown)


archive-r:
If I understand correctly, this would also help with the batch process, correct? 
If this is the function of manually passing the previous 30 seconds context, that is. 
Thanks.

faster_whisper/transcribe.py
```diff
@@ -361,9 +362,12 @@ def generate_segments(
         prompt_reset_since = 0
 
         if options.initial_prompt is not None:
-            initial_prompt = " " + options.initial_prompt.strip()
-            initial_prompt_tokens = tokenizer.encode(initial_prompt)
-            all_tokens.extend(initial_prompt_tokens)
+            if isinstance(options.initial_prompt, Iterable):
```
guillaumekln, line 365: I think `str` is also an iterable. To be safe, can you swap the if-else and check for `str` first?

guillaumekln (changes requested):

faster_whisper/transcribe.py
FlippFuzz: Oops. you are correct. Making the change!

guillaumekln (approved):

PR #356, opened 2023-07-17
Return result with best log prob when all temperature fallbacks failed

archive-r:
Related to https://github.com/openai/whisper/pull/1377

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -671,9 +671,30 @@ class WhisperModel:
                 break
         else:
             # all failed
-            result, avg_logprob, final_temperature, compression_ratio = max(
-                results.values(), key=lambda r: r[1]
-            )
+            filtered_results = {}
+            avg_logprob_index = 1
+            compression_ratio_index = 3
+
+            # filter out results with compression ratio below compression_ratio_threshold
+            for key, result_data in results.items():
+                if (
+                    options.no_speech_threshold is not None
+                    and result_data[compression_ratio_index]
+                    < options.compression_ratio_threshold
+                ):
+                    filtered_results[key] = result_data
+
+            # select the optimal result based on the maximum avg_logprob
+            if filtered_results:
+                result, avg_logprob, final_temperature, compression_ratio = max(
+                    filtered_results.values(),
+                    key=lambda result_data: result_data[avg_logprob_index],
+                )
+            else:
+                result, avg_logprob, final_temperature, compression_ratio = max(
+                    results.values(),
+                    key=lambda result_data: result_data[avg_logprob_index],
+                )
 
         return result, avg_logprob, final_temperature, compression_ratio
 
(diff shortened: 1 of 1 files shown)


archive-r:
Apply the improvements suggested in https://github.com/openai/whisper/pull/1377#issuecomment-1641696612

Hallucinations were noticeably reduced and overall transcription quality increased.

faster_whisper/transcribe.py
```diff
@@ -661,6 +669,32 @@ def generate_with_fallback(
...
+                if (
+                    options.no_speech_threshold is not None
+                    and result_data[compression_ratio_index]
+                    < options.compression_ratio_threshold
+                ):
+                    filtered_results[key] = result_data
+
+            # select the optimal result based on the maximum avg_logprob
+            if filtered_results:
+                result, avg_logprob, final_temperature, compression_ratio = max(
```
guillaumekln, line 689: Let's avoid unpacking the values to avoid repeating the variables multiple times.

Similar to the function `decode_with_fallback` in openai/whisper, you can save the result in a tuple named for example `decode_result` and return this tuple directly.

guillaumekln (commented):
Since this PR only changes the temperature fallback final result, which would be very different than openai/whisper anyway, we could merge it before it is accepted is openai/whisper.

Here are some comments:

archive-r:
Thank you, I have changed the code according to your advice.

faster_whisper/transcribe.py
```diff
@@ -661,8 +672,14 @@ def generate_with_fallback(
 
             if not needs_fallback:
                 break
+        else:
+            # all failed, select the result with the highest log prob
+            if below_cr_threshold_results:
+                decode_result = max(below_cr_threshold_results, key=lambda x: x[3])
+            else:
+                decode_result = max(all_results, key=lambda x: x[3])
```
guillaumekln, line 680: The index should be 1.

You can write this in a single line (not sure if it is too long for the current style check):

```suggestion
            decode_result = max(below_cr_threshold_results or all_results, key=lambda x: x[1])
```

faster_whisper/transcribe.py
```diff
@@ -587,6 +587,9 @@ def generate_with_fallback(
             round(options.max_initial_timestamp / self.time_precision)
         )
 
+        all_results = []
+        below_cr_threshold_results = []
+
```
guillaumekln, line 592: Add before the loop:

```python
decode_result = None
```

We can also remove these initialization lines:

```python
result = None
avg_logprob = None
final_temperature = None
compression_ratio = None
```

archive-r:
Thanks and please review again. Also, I think it would be a good idea to change the PR names to be more intuitive.

archive-r:
There seems to be another benefit to making this change.
Before this change, there were quite a few statements that were okay, even if they had avg_logprob < -1.
However, after applying this change, sentences with avg logprob < -1 in decode_result were almost always completely badly transcribed.
Therefore, I was able to filter them out and further reduce the hallucinations.

PR #359, opened 2023-07-18
Remove duplicate code

archive-r:
Both `median_duration` and `max_duration` have already been calculated on lines 719-720.


guillaumekln:
Thanks!

PR #394, opened 2023-08-02
Token repeat limit, hallucination helper option.

bfs15:
Implemented an argument option to transcribe `token_repeat_limit `.
Predictions with token repeats (consecutive) over this, are failed (with fallback, like other thresholds)
Implemented with filtering of simple extreme repetition (100+ tokens) in mind.

By default this option is disabled of course, and doesn't go inside the `if` branch added.

Predictions with token repeats such as this one are failed (has 222 consecutive token repeats, obviously a hallucination):

```python
[Segment(id=1, seek=200, start=0.0, end=2.0, text=' Suggestions!', tokens=[50364, 39131, 2629, 626, 0, 50464], temperature=0.4, avg_logprob=-0.07976385182804531, compression_ratio=16.59259259259259, no_speech_prob=0.30723854899406433, words=None), Segment(id=2, seek=1395, start=2.0, end=13.950000000000001, text=' Hmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm', tokens=[50364, 8239, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174, 2174], temperature=0.2, avg_logprob=-0.056518122487597994, compression_ratio=31.928571428571427, no_speech_prob=0.6861338019371033, words=None)]
```

The correct caption for this audio is:
```
 Do you have any suggestions?
 Hmmm...
 I don't know...
 Like, like...
```
The "Hmmmm..." prediction goes so long for the prediction with many repeats that it eats up the following other dialogue for some reason (segment from 2.0s to 13.9s, when the Hmmm... actually takes around 1.82 seconds).

I've seen similar repeats happen multiple times in my experience with different audio tracks, but if you want the sample I can share.

Kind of specific, but also very simple addition.

A token_repeat_size could be implemented for cases where tokens are repeated like [x,y, x,y, x,y, ... ], instead of just [x,x,x...]

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -20,6 +20,24 @@ from faster_whisper.vad import (
     get_speech_timestamps,
 )
 
+def consecutive_repeat_sizes(arr):
+    result = {}
+    current_num = None
+    current_count = 0
+
+    for num in arr:
+        if num == current_num:
+            current_count += 1
+        else:
+            if current_num is not None:
+                result[current_num] = max(result.get(current_num, 0), current_count)
+            current_num = num
+            current_count = 1
+
+    if current_num is not None:
+        result[current_num] = max(result.get(current_num, 0), current_count)
+
+    return result
 
 class Word(NamedTuple):
     start: float
@@ -61,6 +79,7 @@ class TranscriptionOptions(NamedTuple):
     word_timestamps: bool
     prepend_punctuations: str
     append_punctuations: str
+    token_repeat_limit: int
 
 
 class TranscriptionInfo(NamedTuple):
@@ -182,6 +201,7 @@ class WhisperModel:
         append_punctuations: str = "\"'.。,，!！?？:：”)]}、",
         vad_filter: bool = False,
         vad_parameters: Optional[Union[dict, VadOptions]] = None,
+        token_repeat_limit: int = 100,
     ) -> Tuple[Iterable[Segment], TranscriptionInfo]:
         """Transcribes an input file.
 
@@ -331,6 +351,7 @@ class WhisperModel:
             word_timestamps=word_timestamps,
             prepend_punctuations=prepend_punctuations,
             append_punctuations=append_punctuations,
+            token_repeat_limit=token_repeat_limit,
         )
 
         segments = self.generate_segments(features, tokenizer, options, encoder_output)
@@ -426,6 +447,15 @@ class WhisperModel:
                     continue
 
             tokens = result.sequences_ids[0]
+            counts = consecutive_repeat_sizes(tokens)
+            
+            for k, v in counts.items():
+                if v > options.token_repeat_limit:
+                    self.logger.debug(f"Many token repets, likely hallucination ({v} > {options.token_repeat_limit})")
+                    # fast-forward to the next segment boundary
+                    seek += min(4, segment_size // 10)
+                    encoder_output = None
+                    continue
 
             previous_seek = seek
             current_segments = []
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -247,7 +247,7 @@ class WhisperModel:
             without speech. This step is using the Silero VAD model
             https://github.com/snakers4/silero-vad.
           vad_parameters: Dictionary of Silero VAD parameters or VadOptions class (see available
-            parameters and default values in the class `VadOptions`).
+          token_repeat_limit: Predictions with token repeats (consecutive) over this, are failed
 
         Returns:
           A tuple with:
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -20,6 +20,7 @@ from faster_whisper.vad import (
     get_speech_timestamps,
 )
 
+
 def consecutive_repeat_sizes(arr):
     result = {}
     current_num = None
@@ -38,6 +39,7 @@ def consecutive_repeat_sizes(arr):
         result[current_num] = max(result.get(current_num, 0), current_count)
 
     return result
+
 
 class Word(NamedTuple):
     start: float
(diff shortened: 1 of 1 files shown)


guillaumekln:
`compression_ratio_threshold` already triggers a fallback when there are too many repetitions. Why do we need a new option for that?

bfs15:
> `compression_ratio_threshold` already triggers a fallback when there are too many repetitions. Why do we need a new option for that?

>compression_ratio_threshold: If the gzip compression ratio is above this value, treat as failed.

No wonder I didn't find this option to deal with repetition. Thanks.
Is there an intuitive way to decide on a value for the compression_ratio_threshold, for the example I gave of obviously extreme repetitions (100+)? 

Makes me wonder why a reasonably high value for this isn't the default.

Purfview:
Could you share the audio sample?

bfs15:
https://files.catbox.moe/m8jj9r.wav

tests were with the medium model, it does get the transcription right most of the time.

Would any `compression_ratio_threshold` make sure 100 token repeats aren't the output?

Purfview:
I don't get any "Hmmm....":

With VAD:
```
[00:00.000 --> 00:01.700]  Do you have any suggestions?
```

No VAD:
```
[00:00.000 --> 00:01.920]  Do you have any suggestions?
[00:05.420 --> 00:06.820]  I don't know...
[00:06.820 --> 00:07.580]  Like...like...
[00:07.580 --> 00:08.320]  Fuwawa desu!
[00:08.420 --> 00:09.180]  Wa wa wa wa!
[00:09.520 --> 00:10.260]  Mokoko desu!
[00:10.400 --> 00:11.160]  Co co co co!
[00:11.560 --> 00:12.260]  Fuwa mo ko desu!
[00:12.400 --> 00:13.020]  Wa wa wa wa!
```

I used my cli version, it has different defaults.

archive-r:
> I didn't set `compression_ratio_threshold` but it let that `compression_ratio=31` pass...? (and others on repeated tests) Is something else making it ignore the `compression_ratio_threshold`? On my tests it happened on both 0 and non-0 temp segments (0.0, 0.2).
> 
> Tested again to make sure you can see in TranscriptionInfo printed below `compression_ratio_threshold=2.4`, but the result had `compression_ratio=31`. Maybe I'm misunderstanding. (as the normal looking "Do you have any suggestions?" segment pred has compression_ratio=10)


If you set the compression_ratio_threshold to 2.4, fallback will be triggered for segments that exceed it. And after fallback, it will try to choose the best result. If a sentence with a `compression_ratio` of 30 looks like the "best" result, it will be chosen. 

So if you want to limit the results based on `compression_ratio`:

1. manually filter sentences from the final results

2. try this alternative, which will probably likely to give you better results.
https://github.com/guillaumekln/faster-whisper/discussions/371

3. this PR will set `avg_logprob` and `compression_ratio` limits for the final result based on `log_prob_threshold` and `compression_ratio_threshold`.
https://github.com/guillaumekln/faster-whisper/discussions/371#discussioncomment-6540179

bfs15:
> If a sentence with a `compression_ratio` of 31 looks like the "best" result, that will be selected, even if it has a very high `compression_ratio` value. The same is true for `log_prob_threshold`.

gah, and there's no way to override this I suppose. that's a bit sad. thanks

> I don't get any "Hmmm....":

I does only happen rarely, as I said it works *most* of the time; I was unlucky to get it too many times for my liking

archive-r:
Changing the `length_penalty` value can help, but it can affect the result.
https://github.com/guillaumekln/faster-whisper/discussions/371 You can also try this.

PR #395, opened 2023-08-02
Add a threshold_limit_ratio parameter

archive-r:
We can already set the `avg_logprob_threshold` and `compression_ratio_threshold` to trigger the fallback.

However, we still had to rely on the user to filter out segments with very low `avg_logprob` or very high `compression_ratio` values.

Segments with a very low `avg_logprob` or a very high `compression ratio` are very likely to be mis-transcribed or hallucinations, and thus may adversely affect subsequent transcription due to the `condition_on_previous_text` option.

This PR prevents these sentences from being output, and also from being added to the previous token. 
This will be useful for users who want to get better accuracy in exchange for a few missing sentences.

In my use case, a value of `threshold_limit_ratio = 2` worked well when `log_prob_threshold` and `compression_ratio_threshold` were at their defaults, and I was able to reduce the hallucinations considerably. 
However, given the variety of use cases, I would recommend leaving the defaults at `None` and letting users adjust them themselves.

For `no_speech_prob`, I didn't consider this value because it's often inaccurate.

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -50,6 +50,7 @@ class TranscriptionOptions(NamedTuple):
     log_prob_threshold: Optional[float]
     no_speech_threshold: Optional[float]
     compression_ratio_threshold: Optional[float]
+    threshold_limit_ratio: Optional[float]
     condition_on_previous_text: bool
     temperatures: List[float]
     initial_prompt: Optional[Union[str, Iterable[int]]]
@@ -170,6 +171,7 @@ class WhisperModel:
         compression_ratio_threshold: Optional[float] = 2.4,
         log_prob_threshold: Optional[float] = -1.0,
         no_speech_threshold: Optional[float] = 0.6,
+        threshold_limit_ratio: Optional[float] = None,
         condition_on_previous_text: bool = True,
         initial_prompt: Optional[Union[str, Iterable[int]]] = None,
         prefix: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


archive-r:
Or maybe it would be better to allow us to set individual limits for avg_logprob, compression_ratio.

Purfview:
I'm not sure if anything can be improved on these "pseudo VAD" thresholds.

Maybe to achieve your desired effect would be better to change 0.5 to 0 at this line:
https://github.com/guillaumekln/faster-whisper/blob/5c17de17713f65929c7c33add3a9735ff75a945c/faster_whisper/transcribe.py#L562C64-L562C64

archive-r:
Thanks for the feedback.

In my opinion, resetting the prompt when temp > 0 may not be necessary, since the sentences selected after fallback were not always transcribed incorrectly. Of course, it might be helpful, I can't say for sure.

Since `avg_logprob `and `compression_ratio` are used when deciding whether to fallback, I thought they would be relatively more reliable values, and in my case they did indeed correlate.

Purfview:
I meant it as not an additional tweak, but "reset the prompt when temp > 0"  instead of adding new "threshold_limit_ratio" argument.

guillaumekln:
> Maybe to achieve your desired effect would be better to change 0.5 to 0 at this line:
https://github.com/guillaumekln/faster-whisper/blob/5c17de17713f65929c7c33add3a9735ff75a945c/faster_whisper/transcribe.py#L562C64-L562C64

Anyone wants to send a PR making this value configurable with an option? Maybe something like `clear_previous_text_on_temperature`?

Even if the change is not really helpful, it is better than leaving a hardcoded constant in the code.

archive-r:
It seems that what we are trying to improve in this PR can be improved to some extent by https://github.com/guillaumekln/faster-whisper/commit/1a1eb1a027c0bccabd86759b9772509f9c29f9dc.

Also, this PR, which prevents outputting results that exceed a threshold by more than a certain factor and resets the prompt, will only be useful in a small number of cases. 
So I'm closing this PR.

PR #398, opened 2023-08-03
Rename clear_previous_text_on_temperature argument

Purfview:
`prompt_reset_on_temperature` is more clear what it does.


guillaumekln:
I’m ok with that. @archive-r What do you think?

archive-r:
Sure, I agree.

PR #403, opened 2023-08-05
added repetition_penalty to TranscriptionOptions

bfs15:
CTranslate2 has a `repetition_penalty` argument in its model.generate method. It was however missing from WhisperModel.transcribe

Simply added the `repetition_penalty` argument

As it is, after each generation, the output is filtered if it's too repetitive (based on compression_ratio).
`repetition_penalty` can make it so repetition is penalized during the generation (decoding) itself, making some failure cases like I found disappear.

By "failure" in this case I mean taking 23 seconds to process a 13 second clip (which usually takes ~2.4 seconds) because of obvious repetition in multiple fallbacks, as seen in this image  https://files.catbox.moe/d0psud.png, where each line starting with `needs_fallback` shows the result of a `model.generation`.

`repetition_penalty` solves this hallucination problem I was having #394 almost perfectly, (using `repetition_penalty=1.2`).

Just by using `repetition_penalty=1.2` it gets a non-hallucinated reply on the first try, though lower values already help on avoiding the worst cases like above. (Maybe setting `repetition_penalty` above one (always) is undesirable, I could see an implementation where a big hallucination is detected (the first generation has `compression_ratio=10` after all!) and increasing the repetition penalty on the second generation.)

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -47,6 +47,7 @@ class TranscriptionOptions(NamedTuple):
     best_of: int
     patience: float
     length_penalty: float
+    repetition_penalty: float
     log_prob_threshold: Optional[float]
     no_speech_threshold: Optional[float]
     compression_ratio_threshold: Optional[float]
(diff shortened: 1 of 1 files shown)


chainyo:
Do you know how this interacts with actual word repetition in an audio file? @bfs15 @guillaumekln

PR #445, opened 2023-08-30
Added audio duration after VAD to TranscriptionInfo object

MinorJinx:
Simple change that adds the seconds of audio removed by the VAD to the returned info object. This was added next to the duration of the original audio. Silero VAD refers to that as silence so that is the name I used.

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -69,6 +69,7 @@ class TranscriptionInfo(NamedTuple):
     language: str
     language_probability: float
     duration: float
+    silence: float
     all_language_probs: Optional[List[Tuple[str, float]]]
     transcription_options: TranscriptionOptions
     vad_options: VadOptions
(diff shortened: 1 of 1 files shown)


Purfview:
You can see removed audio with DEBUG logger:

```
Processing audio with duration 05:05.877

VAD filter removed 01:06.144 of audio
VAD filter kept the following audio segments: [00:00.000 -> 00:03.972], [00:07.260 -> 00:39.972]......
```

MinorJinx:
That's true, but my goal was to create a way to easily retrieve that value similar to how the duration can be accessed.

guillaumekln:
I understand that Silero VAD refers to these segments as silence. However, the property considered alone makes it sound that it returns the precise silence duration in the audio, while this duration depends on the various VAD parameters.

What do you think about adding a property such as "duration_after_vad" to clarify that it is directly related to the VAD filter?

guillaumekln (approved):

PR #467, opened 2023-09-12
added content_frames field

jhj0517:
Hi, I really appreciate for implementing so much efficient whisper. 
This PR adds `content_frames` to `Segment`, so user can track the progress of the generator without exhausting it.
I personally wish there were `content_frames` for the `Segment`.

This is useful when you need to track progress in other frameworks like `gradio`. 

For example, you can do it like this:
```
import gradio as gr 

def transcribe(**params, progress= gr.Progress()):
    generator, info = faster_whisper_model.transcribe(**params)
    for segment in generator:
        progress(segment.seek / segment.content_frames, desc="Transcribing..")
```

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -31,6 +31,7 @@ class Word(NamedTuple):
 class Segment(NamedTuple):
     id: int
     seek: int
+    content_frames: int
     start: float
     end: float
     text: str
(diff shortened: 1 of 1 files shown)


guillaumekln:
Hi,

I think you can achieve the same thing with `segment.start / info.duration`. Can you check?

jhj0517:
OK, I could acheive the same thing
```
print(f"duration : {info.duration} content_frames: {segment.content_frames}")
# duration : 30.1666875 content_frames: 3016
```

All I have to do is 
`int(info.duration* 100)`

Thank you very much and sorry for my confusion! closed.

PR #487, opened 2023-09-23
Adding speech_chunks and sampling_rate to TranscriptionInfo Class

guilhermehge:
I simply added the result of speech_chunks from sileroVAD and sampling rate from feature_extractor to the class TranscriptionInfo so it can be accessed after performing the transcription. This makes it easier to access those informations and allows us to detect the silent segments with a simple function.

I didn't change much, only added those.


guilhermehge:
I missclicked on "Close pull request" once.

PR #499, opened 2023-09-30
Change onnxruntime requirement to gpu version and update VAD to run on gpu

thomasmol:
See discussions here: https://github.com/pyannote/pyannote-audio/issues/1481, https://github.com/guillaumekln/faster-whisper/issues/493, https://github.com/guillaumekln/faster-whisper/issues/364#issuecomment-1645272083

This pull requests lets the VAD run on gpu using `onnxruntime-gpu` rather than `onnxruntime`. There are some issues when depending on both packages: it will default to the cpu version if both are installed. This is mostly a problem when running faster-whisper in conjunction with [pyannote.audio](https://github.com/pyannote/pyannote-audio) (or other libraries that specifically need to run on gpu using `onnxruntime-gpu`).

### DIFF
--- a/faster_whisper/vad.py
+++ b/faster_whisper/vad.py
@@ -251,13 +251,13 @@ class SileroVADModel:
             ) from e
 
         opts = onnxruntime.SessionOptions()
-        opts.inter_op_num_threads = 1
-        opts.intra_op_num_threads = 1
         opts.log_severity_level = 4
+        opts.graph_optimization_level = onnxruntime.GraphOptimizationLevel.ORT_ENABLE_BASIC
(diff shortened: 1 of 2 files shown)


celliso1:
Does onnxruntime-gpu fall back to CPU support if there is no GPU? Not everyone is using CUDA, some are using CPU.

thomasmol:
Unfortunately it does not, so I don't think this pull request will get accepted. I'll leave it up for now if anyone runs into the same issue I experienced that led me to create this pull request.

Purfview:
FYI, you shouldn't run it on CUDA as the model is not meant to run on it.

Benchmark on ~2h audio with RTX4090:

CUDA: 72.22 seconds
CPU: 15.15 seconds

PR #542, opened 2023-11-04
Adding timestamping information

Noah-Grimaldi:
I've needed to collect timestamp information and now I have added it to the segment information.


MahmoudAshraf97:
Timestamps information cannot be collected before transcription, you need to transcribe first

PR #548, opened 2023-11-07
feat: code for whisper-large-v3

stillmatic:
I think these are the necessary changes

- make `feature_size`/`num_mels` configurable - loaded from `preprocessor_config.json`, which comes from the HF repos
- adds HF large-v3  conversion from @bungerr
- add yue cantonese token
- bumps version to 0.10.0

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -1,8 +1,8 @@
 import itertools
+import json
 import logging
 import os
 import zlib
-import json
 
 from typing import BinaryIO, Iterable, List, NamedTuple, Optional, Tuple, Union
 
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -146,15 +146,25 @@ class WhisperModel:
                 "openai/whisper-tiny" + ("" if self.model.is_multilingual else ".en")
             )
         
-        config_file = os.path.join(model_path, "config.json")
-        if os.path.isfile(config_file):
-            with open(config_file, "r") as config:
-                config = json.load(config)
-                self.feature_size = config.get("num_mel_bins", 80)
+        feature_extractor_file = os.path.join(model_path, "preprocessor_config.json")
+        if os.path.isfile(feature_extractor_file):
+            with open(feature_extractor_file, "r") as f:
+                config = json.load(f)
+            feat_kwargs = {
+                k: config[k]
+                for k in [
+                    "n_fft",
+                    "hop_length",
+                    "feature_size",
+                    "sampling_rate",
+                    "chunk_length",
+                ]
+                if k in config
+            }
+        else:
+            feat_kwargs = {}
 
-        self.feature_extractor = FeatureExtractor(
-            feature_size=self.feature_size,
-        )
+        self.feature_extractor = FeatureExtractor(**feat_kwargs)
         self.num_samples_per_token = self.feature_extractor.hop_length * 2
         self.frames_per_second = (
             self.feature_extractor.sampling_rate // self.feature_extractor.hop_length
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -2,6 +2,7 @@ import itertools
 import logging
 import os
 import zlib
+import json
 
 from typing import BinaryIO, Iterable, List, NamedTuple, Optional, Tuple, Union
 
@@ -87,8 +88,6 @@ class WhisperModel:
         num_workers: int = 1,
         download_root: Optional[str] = None,
         local_files_only: bool = False,
-        feature_size: Optional[int] = None,
-        num_languages: Optional[int] = None,
     ):
         """Initializes the Whisper model.
 
@@ -146,26 +145,15 @@ class WhisperModel:
             self.hf_tokenizer = tokenizers.Tokenizer.from_pretrained(
                 "openai/whisper-tiny" + ("" if self.model.is_multilingual else ".en")
             )
-        # large-v3 uses 128, others use 80
-        # if user explicitly sets n_mels, use that
-        if not feature_size:
-            if "large-v3" in model_size_or_path:
-                feature_size = 128
-            else:
-                feature_size = 80
-
-        # this is hacky - should reference directly from vocab
-        # https://github.com/openai/whisper/commit/c5d42560760a05584c1c79546a098287e5a771eb#diff-ad4e95c840033b439466680ea7a88aa513d47c4acb418364f1d3967ca817d440R277
-        if not num_languages:
-            if "large-v3" in model_size_or_path:
-                num_languages = 100
-            else:
-                num_languages = 99
-
-        self.num_languages = num_languages
+        
+        config_file = os.path.join(model_path, "config.json")
+        if os.path.isfile(config_file):
+            with open(config_file, "r") as config:
+                config = json.load(config)
+                self.feature_size = config.get("num_mel_bins", 80)
 
         self.feature_extractor = FeatureExtractor(
-            feature_size=feature_size,
+            feature_size=self.feature_size,
         )
         self.num_samples_per_token = self.feature_extractor.hop_length * 2
         self.frames_per_second = (
@@ -358,7 +346,6 @@ class WhisperModel:
             self.model.is_multilingual,
             task=task,
             language=language,
-            num_languages=self.num_languages,
         )
 
         options = TranscriptionOptions(
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -87,13 +87,15 @@ class WhisperModel:
         num_workers: int = 1,
         download_root: Optional[str] = None,
         local_files_only: bool = False,
+        feature_size: Optional[int] = None,
+        num_languages: Optional[int] = None,
     ):
         """Initializes the Whisper model.
 
         Args:
           model_size_or_path: Size of the model to use (tiny, tiny.en, base, base.en,
-            small, small.en, medium, medium.en, large-v1, large-v2, or large), a path to a converted
-            model directory, or a CTranslate2-converted Whisper model ID from the Hugging Face Hub.
+            small, small.en, medium, medium.en, large-v1, large-v2, large-v3, or large), a path to a
+            converted model directory, or a CTranslate2-converted Whisper model ID from the HF Hub.
             When a size or a model ID is configured, the converted model is downloaded
             from the Hugging Face Hub.
           device: Device to use for computation ("cpu", "cuda", "auto").
@@ -113,6 +115,9 @@ class WhisperModel:
             are saved in the standard Hugging Face cache directory.
           local_files_only:  If True, avoid downloading the file and return the path to the
             local cached file if it exists.
+          feature_size: Number of mel filters to use for feature extraction. If not set,
+            the number of mel filters is inferred from the model version. The first release
+            used 80 bins, but the large-v3 model uses 128 bins.
         """
         self.logger = get_logger()
 
@@ -141,8 +146,27 @@ class WhisperModel:
             self.hf_tokenizer = tokenizers.Tokenizer.from_pretrained(
                 "openai/whisper-tiny" + ("" if self.model.is_multilingual else ".en")
             )
+        # large-v3 uses 128, others use 80
+        # if user explicitly sets n_mels, use that
+        if not feature_size:
+            if "large-v3" in model_size_or_path:
+                feature_size = 128
+            else:
+                feature_size = 80
 
-        self.feature_extractor = FeatureExtractor()
+        # this is hacky - should reference directly from vocab
+        # https://github.com/openai/whisper/commit/c5d42560760a05584c1c79546a098287e5a771eb#diff-ad4e95c840033b439466680ea7a88aa513d47c4acb418364f1d3967ca817d440R277
+        if not num_languages:
+            if "large-v3" in model_size_or_path:
+                num_languages = 100
+            else:
+                num_languages = 99
+
+        self.num_languages = num_languages
+
+        self.feature_extractor = FeatureExtractor(
+            feature_size=feature_size,
+        )
         self.num_samples_per_token = self.feature_extractor.hop_length * 2
         self.frames_per_second = (
             self.feature_extractor.sampling_rate // self.feature_extractor.hop_length
@@ -334,6 +358,7 @@ class WhisperModel:
             self.model.is_multilingual,
             task=task,
             language=language,
+            num_languages=self.num_languages,
         )
 
         options = TranscriptionOptions(
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -145,7 +145,7 @@ class WhisperModel:
             self.hf_tokenizer = tokenizers.Tokenizer.from_pretrained(
                 "openai/whisper-tiny" + ("" if self.model.is_multilingual else ".en")
             )
-        
+
         feature_extractor_file = os.path.join(model_path, "preprocessor_config.json")
         if os.path.isfile(feature_extractor_file):
             with open(feature_extractor_file, "r") as f:
--- a/faster_whisper/utils.py
+++ b/faster_whisper/utils.py
@@ -77,6 +77,7 @@ def download_model(
 
     allow_patterns = [
         "config.json",
+        "preprocessor_config.json",
         "model.bin",
         "tokenizer.json",
         "vocabulary.*",
--- a/faster_whisper/utils.py
+++ b/faster_whisper/utils.py
@@ -21,6 +21,7 @@ _MODELS = {
     "large-v1": "guillaumekln/faster-whisper-large-v1",
     "large-v2": "guillaumekln/faster-whisper-large-v2",
     "large": "guillaumekln/faster-whisper-large-v2",
+    "large-v3": "gradjitta/ct2-whisper-large-v3",
 }
 
 
@@ -50,7 +51,7 @@ def download_model(
     Args:
       size_or_id: Size of the model to download from https://huggingface.co/guillaumekln
         (tiny, tiny.en, base, base.en, small, small.en medium, medium.en, large-v1, large-v2,
-        large), or a CTranslate2-converted model ID from the Hugging Face Hub
+        large, large-v3), or a CTranslate2-converted model ID from the Hugging Face Hub
         (e.g. guillaumekln/faster-whisper-large-v2).
       output_dir: Directory where the model should be saved. If not set, the model is saved in
         the cache directory.
--- a/faster_whisper/utils.py
+++ b/faster_whisper/utils.py
@@ -21,7 +21,7 @@ _MODELS = {
     "large-v1": "guillaumekln/faster-whisper-large-v1",
     "large-v2": "guillaumekln/faster-whisper-large-v2",
     "large": "guillaumekln/faster-whisper-large-v2",
-    "large-v3": " bababababooey/faster-whisper-large-v3",
+    "large-v3": "bababababooey/faster-whisper-large-v3",
 }
 
 
--- a/faster_whisper/utils.py
+++ b/faster_whisper/utils.py
@@ -21,7 +21,7 @@ _MODELS = {
     "large-v1": "guillaumekln/faster-whisper-large-v1",
     "large-v2": "guillaumekln/faster-whisper-large-v2",
     "large": "guillaumekln/faster-whisper-large-v2",
-    "large-v3": "gradjitta/ct2-whisper-large-v3",
+    "large-v3": " bababababooey/faster-whisper-large-v3",
 }
 
 
--- a/faster_whisper/version.py
+++ b/faster_whisper/version.py
@@ -1,3 +1,3 @@
 """Version information."""
 
-__version__ = "0.9.0"
+__version__ = "0.10.0"
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 av==10.*
-ctranslate2>=3.17,<4
+ctranslate2>=3.21,<4
 huggingface_hub>=0.13
 tokenizers>=0.13,<0.15
 onnxruntime>=1.14,<2
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,24 @@ venv/
 # Ignore IDE, Editor Files
 .idea/
 .vscode/
+
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ pip install --force-reinstall "faster-whisper @ https://github.com/guillaumekln/
 ```python
 from faster_whisper import WhisperModel
 
-model_size = "large-v2"
+model_size = "large-v3"
 
 # Run on GPU with FP16
 model = WhisperModel(model_size, device="cuda", compute_type="float16")
--- a/faster_whisper/tokenizer.py
+++ b/faster_whisper/tokenizer.py
@@ -15,7 +15,6 @@ class Tokenizer:
         multilingual: bool,
         task: Optional[str] = None,
         language: Optional[str] = None,
-        num_languages: Optional[int] = 99,
     ):
         self.tokenizer = tokenizer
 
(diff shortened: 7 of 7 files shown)


stillmatic:
fyi @gradjitta -- using your translated model here

Purfview:
Is this enough without [CTranslate2 update](https://github.com/OpenNMT/CTranslate2/pull/1530)?

stillmatic:
> Is this enough without https://github.com/OpenNMT/CTranslate2/pull/1530?

to be honest I'm not sure what that PR does? my read is that it updates some docstrings and adds the `n_mels()` function, but none of the logic changes inside `encode`. in the `faster-whisper` code, the feature extraction is done in Python (and already has an `n_mels` arg) and just passes the extracted array to CTranslate2, which works. 

What would be helpful to add in CTranslate2 is the corresponding `num_languages` function.

funboarder13920:
I don't think that propagating num_languages is necessary as it only matters for one language. We can assume that people forcing the inference to yue know that they should not use large-v2 or large-v1.
large-v3 will be set to the new default large model

yue needs to be added to _LANGUAGE_CODES

n_mels can be infered from the model params, which is why I am updating ctranslate2 to propagate this param.

stillmatic:
> I don't think that propagating num_languages is necessary as it only matters for one language. We can assume that people forcing the inference to yue know that they should not use large-v2 or large-v1.

v3 supports code-switching https://twitter.com/jeremyphoward/status/1721696652506100175 so multi-language support does matter

ThomasKluiters:
When trying to use .transcribe(language='de') an error is returned, as the model is "not multilingual".

stillmatic:
> When trying to use .transcribe(language='de') an error is returned, as the model is "not multilingual".

reproduced, also with another conversion: https://github.com/guillaumekln/faster-whisper/pull/549/files#diff-dff5046df32208d1eaf3b13702e9c9a4c5b44ac2451aa59a6f31aebf8b4d66e3R24

seems like the conversion is tricky to get the `is_multilingual` support (this logic hasn't changed in the PR)

>  You didn't add the new language to _LANGUAGE_CODES so your change has no effect.

https://github.com/guillaumekln/faster-whisper/blob/9c378b681c5c73795e01f7d6aaec368ece6499cf/faster_whisper/tokenizer.py#L278

funboarder13920:
In ctranslate2 is_ultilingual is inferred from the size of the vocabulary, the condition is not met anymore

funboarder13920:
Can you give me precisely where in your PR the num_languages is necessary ?

stillmatic:
you're right - after thinking about it more, I think it has no effect on the current release. however -- if the next version adds another language, then we will be in the same boat. I think it's useful to keep this logic here, I took it from the upstream Python repo and feel like it's more future proof. the `is_multilingual` logic could be similarly handled, using the same logic as upstream

funboarder13920:
In the openai repo they have the language logic because they wrapped the tiktoken tokenizer.
Faster-whisper is using the huggingface tokenizer and does not really handle the language encoding, everything is done in the hugginface tokenizer.
I think there won't be any change in the hugginface tokenizer, it won't require any parameter change either, we just need the new vocabulary file in the download folder and the tokenizer should be v3 compatible.
In the future, if openai releases a new model with different languages, just having the new vocab file and the weights in the ctranslate2 format will be enough.

I agree that is_multilingual is not very robust, especially if a model is release with less languages or with a different amount of tokens. I think that the logic should remain in ctranslate2 but could be improved if another issue comes up in the future.

I will submit a PR when the HF model will be released and after having run some tests.

Purfview:
>  when the HF model will be released

It's now released: https://huggingface.co/openai/whisper-large-v3

bungerr:
i have a working conversion uploaded to bababababooey/faster-whisper-large-v3

yeah the `is_multilingual` check never hits so you'll get something like `The current model is English-only but the language parameter is set to 'yue'; using 'en' instead.` if you try to use another lang. i just kind of lazily commented out all the multilingual checks in `transcribe.py` to test and it didnt break anything

bungerr:
okay i noticed quality with the english only model/tokenizer was significantly worse (missing punctuation, not segmenting sentences, etc) than when forcing the multilingual tokenizer, so i'm using multilingual for everything. seems like im getting the same quality as the openai whisper package now

i think the only thing missing now is getting `is_multilingual` properly set in the ct2 model? it breaks language detection (currently defaults to 'en'). if you try to force it, [it throws a runtimeerror ](https://github.com/OpenNMT/CTranslate2/blob/d0a922737f89888de61196e10ef04069948a8297/python/cpp/whisper.cc#L283)

```
  File "C:\git\faster-whisper-3\faster_whisper\transcribe.py", line 335, in transcribe
    results = self.model.detect_language(encoder_output)[0]
RuntimeError: detect_language can only be called on multilingual models
```

@stillmatic @funboarder13920

flyingleafe:
@stillmatic Opened basically the same PR yesterday lol
I think that it is best to determine `is_multilingual` and `num_languages`, as well as mel parameters, systematically from the configs provided by HF, if those are available (first two are derived from `vocab_size` in `config.json`, the latter are given in `preprocessor_config.json`). The latter part of this is done in my PR, feel free to merge that in here and I will close my one, since yours was the first.

funboarder13920:
The HF model was released but the tokenizer is not in the required format and I am not able to convert it to the fast tokenizer format tokenizer.json 
The pytorch_model.bin is not available either but I was able to convert the model.safetensors to the bin format.

I will wait for the official files. If people are in a rush to use large-v3 they can use any of the proposed PR

RafaRed:
As I cited on [CTranslate2/pull/1530](https://github.com/OpenNMT/CTranslate2/pull/1530#issuecomment-1801506044)

> The current check for multilingual support seems to be hardcoded with a specific vocabulary size:
> `_is_multilingual = vocabulary.size() == 51865;`
> [Link to code](https://github.com/OpenNMT/CTranslate2/blob/d0a922737f89888de61196e10ef04069948a8297/src/models/whisper.cc#L73C3-L73C3)
> 
> For instance, I believe the whisper-latest-v3 model has a vocabulary size of 51866, which is one more than the hardcoded value. This discrepancy could lead to the multilingual feature being incorrectly disabled for this model.
> 
> Probably a more dynamic check need to be implemented to ensure compatibility with future models.

I could make the model work doing the changes suggested by @flyingleafe and removing the last line for the vocabulary.json `"<|30.02|>"` but, of course it will affect the model and it's not a solution.

stillmatic:
cool - let's wait for @funboarder13920 PR to be merged, the multilingual fix might be the only thing necessary there?

the other question is how to handle the `num_languages` and related tokenizer things. I would like to get that from CTranslate2, as that would be consistent with other operations that rely on vocab size (ie the is_multilingual check). 

question on the `preprocessor_config.json` -- can we get it from the `config.json` on the hub that already exists? we're already downloading it here: https://github.com/guillaumekln/faster-whisper/blob/master/faster_whisper/utils.py#L78C10-L78C16 - has `num_mel_bins`. let me try and update to use this file, so that we do not need to download an additional new one.

fyi - pytorch weights are also available on hub: https://huggingface.co/openai/whisper-large-v3/blob/main/pytorch_model.bin

funboarder13920:
I will make a num_languages available in the ctranslate2 model. I don't see when we will need the num_languages though.

stillmatic:
I think we should be able to succeed now just on this ... testing on CTranslate2 head commit now

faster_whisper/utils.py
```diff
@@ -21,6 +21,7 @@
     "large-v1": "guillaumekln/faster-whisper-large-v1",
     "large-v2": "guillaumekln/faster-whisper-large-v2",
     "large": "guillaumekln/faster-whisper-large-v2",
+    "large-v3": "gradjitta/ct2-whisper-large-v3",
```
funboarder13920, line 24: This repo is not official, was not generated from the official HF openai/whisper-large-v3 repo and the tokenizer.json is likely wrong, there is no yue in the languages.

stillmatic: which would you suggest using? @bungerr has one at `bababababooey/faster-whisper-large-v3` but didn't copy the `preprocessor_config.json` over, which is useful to get the feature extractor setup correctly with.

bungerr: oop i just added it in, had to copy the alignment heads over from there & forgot

also accidentally wiped my local openai/whisper-large-v3 repo, pulling again & will reconvert when the new wheels finish building https://github.com/OpenNMT/CTranslate2/actions/runs/6801730648

bungerr: note: tokenizer.json is not provided in the official repo so if you want to run the ct2 translation script, you'll have to either use hf's openai_to_hf converter script (which may have been updated since i last checked(?)), or just do what i did and load+save from AutoTokenizer

```
from transformers import AutoTokenizer

tokenizer = AutoTokenizer.from_pretrained("openai/whisper-large-v3")
tokenizer.save_pretrained("./tokenizerjson")
```

bungerr:
thanks @funboarder13920! after installing the wheel from https://github.com/OpenNMT/CTranslate2/actions/runs/6801730648, `is_multilingual` is properly set & language detection works as expected

i fixed an off-by-1 bug in `config.json` in `bababababooey/faster-whisper-large-v3`, the `<|en|>` token was missing from `"lang_ids"` and `"suppress_ids_begin"` was off by 1 as well

i got tokenizer.json from my method above. i spotted the bug by comparing with the tokenizer.json from [HF's conversion script,](https://huggingface.co/versae/whisper-large-v3) but it seems to be missing the `<|yue|>` token, causing similar off-by-1 issues down the line (ex, all `transcribe` tasks become `translate` tasks)

stillmatic:
cool, I've updated to use @bungerr 's HF repo. installing CTranslate2 from scratch is a bit annoying, would be good to get a release cut there. does the PR work now with the updated repo?

faster_whisper/utils.py
```diff
@@ -21,7 +21,7 @@
     "large-v1": "guillaumekln/faster-whisper-large-v1",
     "large-v2": "guillaumekln/faster-whisper-large-v2",
     "large": "guillaumekln/faster-whisper-large-v2",
-    "large-v3": "gradjitta/ct2-whisper-large-v3",
+    "large-v3": " bababababooey/faster-whisper-large-v3",
```
bungerr, line 24: the leading space breaks the download

bungerr:
haha yeah i was never able to build ct2 from scratch. i think this is the closest thing to a release until a new version gets published to pypi, just have to find the right wheel for your platform+python version & pip install. maybe we should do this in requirements.txt if the release never comes

and yes, i just pulled the new version and is working fine as far as i can tell

stillmatic:
I've opened https://github.com/OpenNMT/CTranslate2/issues/1532 to ask for a new release

AvivSham:
@Purfview Can you help promote this PR to be merged? In addition, I think that due to @guillaumekln limited time, we need new / co maintainer.

ThomasKluiters:
I have tried the models, but I'm having some strange issues where punctuation is definitely no acting the same, will keep testing...

EuropeDev (approved):

blackpolarz:
On my end, I tried the v3 model and noticed that its significantly slower and worse translations than v2. VAD also seems to be cutting the words at weird parts. I will test more and see how it goes.

stillmatic:
I think the changes should wait until the CTranslate2 release is cut, otherwise it will be a confusing experience -- the multilingual tokenizer issue means that results will silently degrade.

stillmatic:
(will rerun CI when the ct2 wheels get uploaded)

stillmatic:
hi @nguyendc-systran -- think this is ready for review!

iorilu:
any one can merge this ?
or only wait Guillaume Klein to do this?

funboarder13920:
Did you benchmark openai vs this PR implementation ?
It seems that there are some errors in the HF hub whisper-large-v3, another PR for faster-whisper mentions that the tokenizer must be different https://github.com/guillaumekln/faster-whisper/pull/559 , if they are right.

funboarder13920:
I am still waiting for the large-v3 HF hub to be aligned with all the other models, specifically the tokenizer.json config before trying anything

flyingleafe:
@funboarder13920 There is a very simple way to obtain the `tokenizer.json` file from the HF model.
```
from transformers import WhisperTokenizer, WhisperTokenizerFast

slow = WhisperTokenizer.from_pretrained("openai/whisper-large-v3")
fast = WhisperTokenizerFast(__slow_tokenizer=slow)
fast.save_pretrained("/path/to/dir")
```

This yields all the necessary files for the tokenizer _including_ `tokenizer.json`

flyingleafe:
@funboarder13920 alternatively you can use conversion script from my PR to transformers heh (https://github.com/huggingface/transformers/pull/27336)

I pinged Sanchit there though regarding `tokenizer.json` in the official checkpoint to save everybody some hassle, so I think the issue should be resolved soon.

Purfview:
@flyingleafe what about `config.json` issue mentioned there -> https://github.com/guillaumekln/faster-whisper/pull/548#issuecomment-1802817971

flyingleafe:
@Purfview I figured the issue out, so basically the difference between `large-v2` and `large-v3` tokenizer checkpoints is that `<|endoftext|>` token is included in the additional special tokens list for previous checkpoints versions, but excluded from it for `large-v3`.

This token is BOS token so by the logic of `tokenizers` library it should not be in this list in the first place, so the fact it was there before is some kind of bug apparently. But CTranslate2 converter relied on this token being present in the list.

I opened the PR in CTranslate2 fixing this: https://github.com/OpenNMT/CTranslate2/pull/1546

The other issue which was present in the converter is that they did not get the suppressed token values from HF model's `generation_config`. Those are important for proper transcription (all the faster-whisper checkpoints by @guillaumekln so far included those, and they were equivalent to HF ones). I fixed that too.
https://huggingface.co/flyingleafe/faster-whisper-large-v3 <- this checkpoint is made using those fixes, together with this PR seems to be working well. @ThomasKluiters please feel free to test your punctuation issues on this checkpoint, I suspect those could be due to nonexistent `suppress_tokens` set in your previous attempts.

Purfview:
@flyingleafe I'm not an expert, but your fixes make sense. Are you sure that there are no other quirks with tokens?
Going to test it, I'll report later.

ThomasKluiters:
@flyingleafe Thanks! I'm gonna have a look later today!

Purfview:
@flyingleafe I didn't noticed any difference so far.
Btw, there is new commit: https://huggingface.co/openai/whisper-large-v3/commit/99ef777c8f3c81bade67c14cfa359a3d6303b788

flyingleafe:
@Purfview can you send me some of your test samples maybe? maybe my samples are too easy, so far the results I see are virtually identical to `large-v2`.

the half of the new commit to the HF checkpoint makes sense (bos/eos/pad token ids are wrong indeed), but the reduction of the vocabulary size actually does not make sense, will test HF version today with and without this change to figure out why they did that.

Purfview:
> send me some of your test samples

Check [this](https://we.tl/t-EfsZBzImfj).

funboarder13920:
I run some benchmarks to compare large-v2 with large-v3. large-v2 is currently way better than large-v3.
Some possible explanations:
- large-v3 is not accurate
- there is an issue with the conversion (openai to hf or hf to ct2)
- there is an issue with the code from HF we are using
- there is an issue with this code
- the default inference params are different from large-v2 ?

Next week I will benchmark openai/large-v3 against hf/large-v3 against ct2/large-v3 and openai/large-v2 against openai/large-v3 to narrow down the possibilities.

flyingleafe:
@funboarder13920 can confirm, on the example provided by the @Purfview large-v3 hallucinates a lot of BS in the end and in-between phrases, while large-v2 does not; no problems with punctuation though

iorilu:
after this long discussion 

any update on the final result 

as I want to be if it's ok to update to v3, at least with no negative impact copmare to v2

iorilu:
> I ran some benchmarks to compare large-v2 with large-v3. large-v2 is currently way better than large-v3. Some possible explanations:
> 
> * large-v3 is not accurate
> * there is an issue with the conversion (openai to hf or hf to ct2)
> * there is an issue with the code from HF we are using
> * there is an issue with this code
> * the default inference params are different from large-v2 ?
> 
> Next week I will benchmark openai/large-v3 against hf/large-v3 against ct2/large-v3 and openai/large-v2 against openai/large-v3 to narrow down the possibilities.

what language do you use to test ? test on mulitiple language or just English?

blackpolarz:
From my own testing, the same problem occurs with whisper v3 translation (JP to EN) task. For most parts, whisper v3 has better grammar, but it tends to hallucinate and fill in words by itself, which can cause the context to change at times. This occurs even with the Hugging face's implementation, so I am not sure if the problem lies with the conversion from open-ai's model to hf or the model itself. Whisper v2 is slightly better in terms of "word for word" so context don't change as much but hallucination is still a problem. 
I also noticed that "avg_logprob" in whisper v3 tends to be lower when testing the same audio on whisper v2.

doublex:
@flyingleafe 
Did you use this model in your test?
https://huggingface.co/flyingleafe/faster-whisper-large-v3

faster_whisper/transcribe.py
```diff
@@ -113,6 +114,9 @@ def __init__(
             are saved in the standard Hugging Face cache directory.
           local_files_only:  If True, avoid downloading the file and return the path to the
             local cached file if it exists.
+          feature_size: Number of mel filters to use for feature extraction. If not set,
```
funboarder13920, line 117: Not used anymore

faster_whisper/transcribe.py
```diff
@@ -142,7 +146,25 @@ def __init__(
                 "openai/whisper-tiny" + ("" if self.model.is_multilingual else ".en")
             )
 
-        self.feature_extractor = FeatureExtractor()
+        feature_extractor_file = os.path.join(model_path, "preprocessor_config.json")
```
funboarder13920, line 149: Maybe move that into a specific method ?
And use the n_mels from ct2 as a fallback ?

requirements.txt
```diff
@@ -1,5 +1,5 @@
 av==10.*
-ctranslate2>=3.17,<4
+ctranslate2>=3.21,<4
 huggingface_hub>=0.13
 tokenizers>=0.13,<0.15
```
funboarder13920, line 4: Add support to tokenizer 0.15 version, otherwise it is inconvenient to install faster-whisper and transformers/tokenizers

faster_whisper/version.py
```diff
@@ -1,3 +1,3 @@
 """Version information."""
 
-__version__ = "0.9.0"
```
funboarder13920, line 3: this commit is for the release

faster_whisper/utils.py
```diff
@@ -21,6 +21,7 @@
     "large-v1": "guillaumekln/faster-whisper-large-v1",
     "large-v2": "guillaumekln/faster-whisper-large-v2",
     "large": "guillaumekln/faster-whisper-large-v2",
+    "large-v3": "bababababooey/faster-whisper-large-v3",
```
funboarder13920, line 24: who is owning that ?
Can systran create its hub to keep ownership of the models and easily update old models and upload new models ?

requirements.txt
```diff
@@ -1,5 +1,5 @@
 av==10.*
-ctranslate2>=3.17,<4
+ctranslate2>=3.21,<4
```
nguyendc-systran, line 2: New release 3.22.0 on Ctranslate2 is in progress: https://github.com/OpenNMT/CTranslate2/pull/1561  
Please update this accordingly.

stillmatic: will do, please ping when it's on pypi

requirements.txt
stillmatic: do you know if any compatibility fixes are necessary? or will just `tokenizers>=0.13` suffice?

faster_whisper/utils.py
Purfview: > who is owning that ?

Some user posted that link on issues.

Best would be to keep model under official acc like "systran", maybe move all models there.

nguyendc-systran: Good point, we are waiting for the release on CTranslate2, then push the new converted model (with the fix by https://github.com/OpenNMT/CTranslate2/pull/1546) to Systran organization

faster_whisper/transcribe.py
stillmatic: I'll try to get to this tomorrow. extremely limited bandwidth with American holidays currently

requirements.txt
funboarder13920: I guess Guillaume did that because ctranslate2 relies heavily on the tokenizers, to prevent a new tokenizers version to introduce breaking change into faster-whisper. We can go for tokenizers>=0.13,<0.16

requirements.txt
archive-r: https://github.com/OpenNMT/CTranslate2/releases/tag/v3.22.0

requirements.txt
nguyendc-systran: We made some tests on ourside, the tokenizers 0.15 works normally, so: tokenizers>=0.13,<0.16 looks good for me.

faster_whisper/transcribe.py
```diff
@@ -142,7 +146,25 @@ def __init__(
...
+                config = json.load(f)
+            feat_kwargs = {
+                k: config[k]
+                for k in [
+                    "n_fft",
+                    "hop_length",
+                    "feature_size",
+                    "sampling_rate",
+                    "chunk_length",
+                ]
```
nguyendc-systran, line 161: Minor remark: in this new method could you make these parameters less "hard-code" way because they come from this class: https://github.com/SYSTRAN/faster-whisper/blob/master/faster_whisper/feature_extractor.py#L8-L12

faster_whisper/utils.py
Oscaarjs: @nguyendc-systran thank you! So essentially just waiting for @stillmatic to do the last fixes you mentioned before merge?

Purfview: @flyingleafe 
Why tokenizer.json & vocabulary.json is not he same in https://huggingface.co/flyingleafe/faster-whisper-large-v3 vs your https://huggingface.co/flyingleafe/faster-whisper-large-v3 ?

AvivSham: @Purfview I think you accidentally pasted the same URL.

Purfview: @AvivSham Fixed it.

blackpolarz: Not sure if I missed out any but in tokenizer.json, there is a difference in token 50363, nospeech vs nocaptions.

Oscaarjs:
@nguyendc-systran in [https://huggingface.co/Systran/faster-whisper-large-v3/tree/main](url) `vocabulary` is a `.json` but for the other models e.g. [https://huggingface.co/Systran/faster-whisper-large-v2/tree/main](url) its a `.txt` file. Is this intended?

faster_whisper/utils.py
nguyendc-systran: IMHO, yes.
Another point may be interesting / relevant is about this benchmark: https://github.com/SYSTRAN/faster-whisper/pull/548#issuecomment-1814884183.   
Not sure if @funboarder13920 had a chance for that?

minhthuc2502:
> @nguyendc-systran in [https://huggingface.co/Systran/faster-whisper-large-v3/tree/main](url) `vocabulary` is a `.json` but for the other models e.g. [https://huggingface.co/Systran/faster-whisper-large-v2/tree/main](url) its a `.txt` file. Is this intended?

Hello, Ct2 support vocabulary in both format ``json`` and ``txt``. Recently, after this [PR](https://github.com/OpenNMT/CTranslate2/pull/1293) all the conversion will generate ``vocabulary.json`` instead of txt. That's why in [https://huggingface.co/Systran/faster-whisper-large-v3/tree/main](https://github.com/SYSTRAN/faster-whisper/pull/url), we have vocabulary.json

faster_whisper/utils.py
funboarder13920: Yep, the inference between openai/hf/faster_whisper are not exactly the same but very similar. I guess I was witnessing the differences between v2 and v3.

Oscaarjs:
Added some changes and reflections from this thread in a new PR #578

nguyendc-systran:
Hi @stillmatic, @Oscaarjs et al.
Thank you very much for the contribution and all discussion in this thread.
I take the liberty of closing this discussion as the remaining works has been done and merged by https://github.com/SYSTRAN/faster-whisper/pull/578
Thanks both of you ;)

Purfview:
Still, nospeech vs nocaptions difference in token 50363 wasn't answered.

PR #554, opened 2023-11-10
Add GUI faster-whisper project README.md

JuergenFleiss:
Added aTrain GUI faster-whisper transcription and diarization tool as community project.


gvijqb:
Hi @Juergen-J-F we would love to contribute if you are making a public project for community. 
We have a scalable API for ASR which uses faster-whisper and has diarization built in as well. You can try here:
https://monsterapi.ai/playground/speech2text

PR #557, opened 2023-11-11
[feat] update distil-whisper

metame-none:
reference issue: https://github.com/guillaumekln/faster-whisper/issues/533

features: `allowing the user to specify the chunk length and also the maximum generation length`

tested ok with latest [ctranslate2](https://github.com/OpenNMT/CTranslate2/commit/1e37b526d1d5b61b132ff6dc144136d06e00fd7c) converted model(upload to huggingface soon)


Purfview:
@nguyendc-systran , would be great if you could have a look.

trungkienbkhn:
@metame-none , hello. How can you generate the ct2 conversion model from hf [distil-whisper](https://huggingface.co/distil-whisper/distil-large-v2) model ?
I tried to convert the distil-whisper model from Hugging face hub to ctranslate2 but the hf-to-ct2 script returns an error. More detail can be view in [issue-1564](https://github.com/OpenNMT/CTranslate2/issues/1564).
After discussing with @nguyendc-systran and @minhthuc2502, we created a new PR to fix this script [pr-1565](https://github.com/OpenNMT/CTranslate2/pull/1565) in ctranslate2.

Besides, I found that the [config.json](https://huggingface.co/metame/faster-distil-whisper-large-v2/blob/main/config.json#L2) file in your hf model has **alignments_head** field, but it seems it is different from the [config.json](https://huggingface.co/openai/whisper-large-v2/blob/main/generation_config.json#L2) file of the openai whisper large-v2 model ? Can you explain how to generate it ?
Thank in advance.

faster_whisper/transcribe.py
```diff
@@ -623,6 +636,10 @@ def generate_with_fallback(
         max_initial_timestamp_index = int(
             round(options.max_initial_timestamp / self.time_precision)
         )
+        if options.max_new_tokens is not None:
+            max_length = min(self.max_length, len(prompt) + options.max_new_tokens)
```
trungkienbkhn, line 640: Can you re-check this logic ? I think we should use max_length option as input instead of creating new param max_new_tokens.
Example: `max_length = min(self.max_length, options.max_length) if options.max_new_tokens is not None else self.max_length`

nguyendc-systran: We can event simplify like:  
`max_length = options.max_length if options.max_length is not None else self.max_length`
  
And please rebase your branch on the last version of master to avoid conflict.

metame-none:
> ctranslate2

hi, by the time I converted, the version I used was [3.21 for ctranslate](https://github.com/OpenNMT/CTranslate2/commit/1e37b526d1d5b61b132ff6dc144136d06e00fd7c), and the lastest is 3.22 and it seems it doesn't test on distil-whisper.

And for the config.json, the alignments_head follows the logic as [code](https://github.com/OpenNMT/CTranslate2/blob/d963499e6ed496526e68fb149c908cb2da1f0255/python/ctranslate2/converters/transformers.py#L929), I also tried to figure out what was the alignments_head for distil-whisper, sadly no luck. But, I tested the word_timestamps for this setting, and it seems working good.

faster_whisper/transcribe.py
metame-none: thx for your advice, I use a new max_new_tokens is just for keeping the same logic as distil-whisper, where the setting is 
```python
pipe = pipeline(
    "automatic-speech-recognition",
    model=model,
    tokenizer=processor.tokenizer,
    feature_extractor=processor.feature_extractor,
    max_new_tokens=128,
    torch_dtype=torch_dtype,
    device=device,
)
```
and the default max_length seems include the length of the prompt. So, to make these two compatable, I introduce a new param, not sure if this is a good way, glad to hear your advice. thx a lot.

trungkienbkhn:
@sanchit-gandhi, tks for your interesting repo, we are supporting to integrate it to faster-whisper. I'd like to clarify the **alignment_heads** field.
For info about aligment_heads:
>To allow the Hugging Face version of Whisper to predict word-level timestamps, a new property alignment_heads must be added to the GenerationConfig object. This is a list of [layer, head] pairs that select the cross-attention heads that are highly correlated to word-level timing.

I found that this field does not exist in the [generation_config.json](https://huggingface.co/distil-whisper/distil-large-v2/blob/main/generation_config.json) file in your HF model, while the original openai model still exists. Could it affect the word-level timing of the transcription results ?

faster_whisper/transcribe.py
metame-none: I remove the new param max_new_tokens, and use the options.max_length instead, thx

metame-none:
I find two issues with faster-distil-whisper:
1.  the running speed of faster-distil-whisper is not that fast as I was expected, especially **sending numpy array** to `result = pipe(sample)` as [link](https://github.com/huggingface/distil-whisper#short-form-transcription), the transformers version seems faster (tested with 3090, roughly **300ms/transformers vs 400ms/faster-distil-whisper** with 10s audio sample)
2. the faster-distil-whisper works worse in the long audio setting, tested with [link](https://youtu.be/KvampKnG1dU) and the  param `condition_on_previous_text` makes a huge impact.

faster_whisper/transcribe.py
```diff
@@ -642,6 +647,10 @@ def generate_with_fallback(
         max_initial_timestamp_index = int(
             round(options.max_initial_timestamp / self.time_precision)
         )
+        if options.max_length is not None:
+            max_length = min(self.max_length, len(prompt) + options.max_length)
```
trungkienbkhn, line 651: @metame-none, hello. Why do you need to get min in this logic ? If we use max_length > default (448 for faster whisper), will it cause too negative effects ? If not, I think can make this logic simpler by removing the min function.

trungkienbkhn:
@metame-none , for https://github.com/SYSTRAN/faster-whisper/pull/557#issuecomment-1832294968:
1. I'm not sure about that. In my tests for both short-form and long-form audio, fw-distilled-whisper is always superior in timing due to the original fw-large-v2.
2. I also confirm this. But if we use ****condition_on_previous_text**=False**, transcription quality has improved quite a lot.

Besides, I think shoud set chunk_length=15 in preprocessor_config.json file:
> To enable chunking, pass the chunk_length_s parameter to the pipeline. For Distil-Whisper, a chunk length of 15-seconds is optimal.

metame-none:
@trungkienbkhn for https://github.com/SYSTRAN/faster-whisper/pull/557#issuecomment-1833291059
1. could you please share your testing code? I would like to re-check with that.
2. Yes, `condition_on_previous_text=False` will imporve a lot.
3. you mean, we should set `chunk_length=15` as the default setting? I just copy it from the origin HF repo.
thx for your time.

trungkienbkhn:
@metame-none,  
1. Below is my testing code:
`model = WhisperModel('/tmp/distil-whisper-large-v2-ct2, device="cuda")`
`segments, info = model.transcribe(jfk_path, word_timestamps=True, condition_on_previous_text=False, max_length=128)`
 I tried converting the large v2 model from the distil-whisper and got a model similar to your HF model.
2. Based on experience, I find that set condition_on_previous_text=False (default is True) will give much better quality. Hope someone can find the reason
3. I think it's best to put chunk_length as an input variable when calling the transcribe function

funboarder13920:
>     2. why fw-distilled is broken with `condition_on_previous_text=True`?

It is likely that the distillation was done without prompting/conditionning on previous text

sanchit-gandhi:
Hey @trungkienbkhn, @metame-none, @Purfview, @funboarder13920! 

First of all, thank you for maintaining this amazing repository! It's such a valuable resource for the open-source speech community 🙌 Super excited to see how it improves over the coming months 🚀

Answering your questions sequentially below:

1. **Alignment heads:** I've not looked at the specific cross-attention plots, but we should be able to use the same alignment heads as in the original `large-v2` (resp. `medium.en`) model for `distil-large-v2` (resp. `distil-medium.en`). _c.f._ https://github.com/openai/whisper/discussions/1733#discussioncomment-7487439
2. **The running speed of faster-distil-whisper is not that fast as I was expected:** have you benchmarked Transformers Whisper vs `faster-whisper`? We're consistently seeing that Transformers Whisper is faster than `faster-whisper`, so it's likely this trend also holds for Transformers Distil-Whisper vs `faster-distil-whisper`. _c.f._ https://github.com/Vaibhavs10/insanely-fast-whisper#insanely-fast-whisper
3. **`condition_on_previous_text`**: the model **is** trained with conditioning on previous context. However, not to the same proportion as the original Whisper model. It's also very hard to maintain the same distribution of previous context tokens as the original Whisper, since we don't have access to the original dataset details 😅 We tried our best to implement this using our intuition, but it might well be that it doesn't perform as well as the original Whisper model with prev text conditioning. Has anyone evaluated the word error rate difference with / without using prev text conditioning? Would be curious to see what the difference is, and possibly provide some insight as to how we can improve this!
4. **Chunk size 15s**: this was set using the Transformers chunking algorithm. For your sequential long-form algorithm, it might well be that 30s is sufficient. This should probably be tuned by evaluating over a corpus of long-form audio and tuning the hyper parameter (as we did for our chunked algorithm)
5. **Instead of distill just use smaller fw model:** Distil-Whisper should be **significantly** more performant than smaller checkpoints, with faster latency! We found this held across the board using our Transformers chunking algorithm.

Purfview:
@sanchit-gandhi 
> we should be able to use the same alignment heads as in the original

I think I tried that and model was outputting empty transcriptions.

> Has anyone evaluated the word error rate difference with / without using prev text conditioning? 

I didn't, but with the prev text conditioning after the first chunk it was outputting non-stop pure hallucinations .

>5... Distil-Whisper should be significantly more performant than smaller checkpoints, with faster latency!

Not the case for faster-whisper. People would want to use distill because it's faster, but this effect it's insignificant with fw. So with smaller fw model you get the wanted faster effect.
Frankly, I just quick tested "distil-medium.en", "fw-small.en", "fw-medium.en".

ostegm:
@sanchit-gandhi Wanted to ask for some clarification here: 

> 2. **The running speed of faster-distil-whisper is not that fast as I was expected:** have you benchmarked Transformers Whisper vs `faster-whisper`? We're consistently seeing that Transformers Whisper is faster than `faster-whisper`, so it's likely this trend also holds for Transformers Distil-Whisper vs `faster-distil-whisper`. _c.f._ https://github.com/Vaibhavs10/insanely-fast-whisper#insanely-fast-whisper

It was a month or two ago, prior to distill whisper, but I tested this and found that the results posted in "insanely fast whisper" are heavily dependent on batching. If you keep batch size at 1 I did not see transformers whisper outperforming faster-whisper. Am I missing something? Are there any other resources confirming that transformers-whisper is indeed faster?

trungkienbkhn:
> we should be able to use the same alignment heads as in the original

FYI, I tested my [distil-whisper conversion model](https://huggingface.co/trungkienbkhn/distil-faster-whisper-large-v2) with the original alignment heads. But I encountered this error:

Screenshot from 2023-12-12 14-33-25

This error only occurs if I set `word_timestamps=True` :
```
model = WhisperModel('distil-large-v2', device='cuda')
segments, info = model.transcribe(jfk_path, condition_on_previous_text=False, word_timestamps=True)
```

After verifying, I found that it was stuck due to this [logic](https://github.com/SYSTRAN/faster-whisper/blob/master/faster_whisper/transcribe.py#L894):
```
result = self.model.align(
            encoder_output,
            tokenizer.sot_sequence,
            [text_tokens],
            num_frames,
            median_filter_width=median_filter_width,
        )[0]
```
Maybe this is a bug on ctranslate2 for the align function.

SinanAkkoyun:
Does anyone have benchmarks on batch size 1 faster distill whisper on 2s audio clips?

metame-none:
> > we should be able to use the same alignment heads as in the original
> 
> FYI, I tested my [distil-whisper conversion model](https://huggingface.co/trungkienbkhn/distil-faster-whisper-large-v2) with the original alignment heads. But I encountered this error:

Hi, @trungkienbkhn, what the alignment heads you use? I tested with the [default alignment heads](https://github.com/OpenNMT/CTranslate2/blob/d963499e6ed496526e68fb149c908cb2da1f0255/python/ctranslate2/converters/transformers.py#L929) and it seems ok。And the origin alignment head for "large-v2" seems starts with layer 10, and the distill-whisper only has two decode layers. 

According to the author of whisper from openai, https://github.com/openai/whisper/discussions/1388#discussioncomment-7568428
> The heads were not specifically designed or constrained to be monotonically aligned, but some heads in the cross attention layers naturally learned to have the attention weights matching the time alignment. The _ALIGNMENT_HEADS masks are obtained post-training by rather manually selecting the heads that have clean alignment patterns.

So I suppose, the default alignment heads should work just fine.

metame-none:
> Does anyone have latency figures on batch size 1 faster distill whisper on 2s audio clips (or at least Ns audio speed comparison between faster whisper and faster distill whisper)

I test with rtx 3090, with 10s audio sample:`faster-distill-large-v2` takes 430ms, `faster-whisper-large-v2` takes 960ms, and `faster-whisper-medium-en` takes 710ms.

SinanAkkoyun:
@metame-none Thank you so so much!!

metame-none:
> 3. I think it's best to put chunk_length as an input variable when calling the transcribe function

hi, @trungkienbkhn , I added the chunk_length as an input of transcribe, thx.

ostegm:
Have you tried this with an initial prompt? I've been trying it out and it seems to break the transcripts when I provide a prompt.

trungkienbkhn:
> what the alignment heads you use?

Hi, @metame-none. I used the original alignment heads starting from layer 10, which caused an error because the decoder only has 2 layers. I agree with you that we should use default alignment heads. Tks.

faster_whisper/transcribe.py
trungkienbkhn: @metame-none, hello.
I think we should rollback to both use max_new_tokens and max_length params, then update this logic similar to what HF transformer is doing:
```python
# If the user passes `max_new_tokens`, increase its number to account for the prompt
if kwargs.get("max_new_tokens", None) is not None:
    kwargs["max_new_tokens"] += len(text_prompt_ids)
    if kwargs["max_new_tokens"] >= self.config.max_target_positions:
        raise ValueError(
            f"The length of the sliced `prompt_ids` is {len(text_prompt_ids)}, and the `max_new_tokens` "
            f"{kwargs['max_new_tokens'] - len(text_prompt_ids)}. Thus, the combined length of the sliced "
            f"`prompt_ids` and `max_new_tokens` is: {kwargs['max_new_tokens']}. This exceeds the "
            f"`max_target_positions` of the Whisper model: {self.config.max_target_positions}. "
            "You should either reduce the length of your prompt, or reduce the value of `max_new_tokens`, "
            f"so that their combined length is less that {self.config.max_target_positions}."
        )
```
=> raise error if max_new_tokens + len(prompt) > max_length
For more info, please check this [pr](https://github.com/huggingface/transformers/pull/26164).

Besides, I also agree that max_length includes the length of the prompt: https://huggingface.co/docs/transformers/main/en/main_classes/text_generation#transformers.GenerationConfig.max_length

Please take a look. Tks.

nguyendc-systran:
@trungkienbkhn please help to generate corresponding Systran models in the HUB.  
@metame-none can you please add a note in the readme regarding impact of the "condition_on_previous_text=False" option when using the distil model?

trungkienbkhn:
@metame-none , hello. FYI, we uploaded the Systran Faster Distil-Whisper conversion models to the HuggingFace hub:
+ [distil-large-v2](https://huggingface.co/Systran/faster-distil-whisper-large-v2)
+ [distil-medium.en](https://huggingface.co/Systran/faster-distil-whisper-medium.en)
+ [distil-small.en](https://huggingface.co/Systran/faster-distil-whisper-small.en)

Please verify then update those models in this project's [utils](https://github.com/SYSTRAN/faster-whisper/blob/master/faster_whisper/utils.py) in your pr.
And could you squad your commit? The logic now looks good to me. Tks.

ostegm:
> Have you tried this with an initial prompt? I've been trying it out and it seems to break the transcripts when I provide a prompt.

HI, just want to bump this comment.

metame-none:
> Have you tried this with an initial prompt? I've been trying it out and it seems to break the transcripts when I provide a prompt.

hi, @ostegm 
emprically, `condition_on_previous_text=True` will degrade the performance for long audio, and I think the initial prompt will likely have the same problem, since there are the same logically.

metame-none:
hi, @trungkienbkhn 
I updated utils.py and readme with Systran models, and test ok.
btw, the alignment heads for the small.en model is wrong in [distil-whisper/distil-small.en](https://huggingface.co/distil-whisper/distil-small.en/blob/main/generation_config.json), so I make a PR in the [distil-small.en](https://huggingface.co/Systran/faster-distil-whisper-small.en). thx

README.md
```diff
@@ -36,6 +38,33 @@ For reference, here's the time and memory usage that are required to transcribe
 
 *Executed with 8 threads on a Intel(R) Xeon(R) Gold 6226R.*
 
+
+### Distil-whisper
+
+| Implementation | Precision | Beam size | Time | Gigaspeech WER |
+| --- | --- | --- | --- | --- |
+| distil-whisper/distil-large-v2 | fp16 | 4 |- | 10.36 |
+| [faster-distil-large-v2](https://huggingface.co/metame/faster-distil-whisper-large-v2) | fp16 | 5 | - | 10.28 |
```
nguyendc-systran, line 47: Can you please update these link according to info in `utils.py`

metame-none: > Can you please update these links according to info in `utils.py`

thx, I updated the readme.

nguyendc-systran: Thanks for your contribution, I merge it now

Issue #621, opened 2023-12-17
Bug: Illogical "Avoid computing higher temperatures on no_speech"


Purfview:
Anyone noticed that instead of silence when no_speech_threshold is triggered it's outputting hallucinations?

```
Processing segment at 00:00.000
DEBUG: Compression ratio threshold is not met with temperature 0.0 (0.846154 > 0.800000)
[00:29.500 --> 00:29.780]  Sottotitoli creati dalla comunità Amara.org
Processing segment at 00:30.000
DEBUG: Compression ratio threshold is not met with temperature 0.0 (0.846154 > 0.800000)
[00:30.000 --> 00:59.980]  Sottotitoli creati dalla comunità Amara.org
```

Respectively from json:
```
            "temperature": 0.0,
            "avg_logprob": -0.37790222728953643,
            "compression_ratio": 0.8461538461538461,
            "no_speech_prob": 0.8777729868888855,
```

```
            "temperature": 0.0,
            "avg_logprob": -0.30678143220789295,
            "compression_ratio": 0.8461538461538461,
            "no_speech_prob": 0.9770684242248535,
```

I've used `compression_ratio_threshold=0.8`.

Purfview:
False alarm, I thought that `compression_ratio_threshold` matters too in "silence", forgot that only `logprob_threshold` is considered.

Purfview:
But wait, if it's not "silence" then why it doesn't go through all fallbacks?

Purfview:
Here it breaks fallbacks because it's "silence":
https://github.com/SYSTRAN/faster-whisper/blob/44f7e589478866546bfcd1d105e254a74e2caad5/faster_whisper/transcribe.py#L721-L725

But here it's not "silence" anymore:
https://github.com/SYSTRAN/faster-whisper/blob/44f7e589478866546bfcd1d105e254a74e2caad5/faster_whisper/transcribe.py#L455-L463

Purfview:
Example of the bug introducing the bad hallucination loop:

```
DEBUG: Compression ratio threshold is not met with temperature 0.0 (6.677966 > 2.400000)
[04:17.320 --> 04:29.020]  been doing it for a long time. I'm a professional. I'm a professional. I'm a
[04:29.020 --> 04:29.340]  professional. I'm a professional. I'm a professional. I'm a professional. I'm
[04:29.340 --> 04:34.560]  a professional. I'm a professional. I'm a professional. I'm a professional. I'm
[...]

PR #625, opened 2023-12-21
Bugfix: Illogical "Avoid computing higher temperatures on no_speech"

Purfview:
Fixing: https://github.com/SYSTRAN/faster-whisper/issues/621
Regression since PR: https://github.com/SYSTRAN/faster-whisper/pull/225

**The bug**: It's "silence" when decoding has failed due to `compression_ratio_threshold` [+no_speech_threshold] in https://github.com/SYSTRAN/faster-whisper/pull/225, when further down the code it's not "silence" anymore.

"Silence" should be only when decoding has failed due to `log_prob_threshold`  [+no_speech_threshold].
As described there:
https://github.com/SYSTRAN/faster-whisper/blob/44f7e589478866546bfcd1d105e254a74e2caad5/faster_whisper/transcribe.py#L239-L241

The bug opens the gates for the hallucination loops.


arabcoders:
@Purfview any reason you decided to close this issue? i am using your patch and it's improving the output result

nguyendc-systran:
Thanks @Purfview for the fix, [this explanation](https://github.com/openai/whisper/pull/1903#issuecomment-1866154976) makes sens for me.
I will merge it now.

PR #646, opened 2024-01-16
Skip silence around hallucinations

trungkienbkhn:
Same with https://github.com/openai/whisper/pull/1838

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -14,7 +14,7 @@ import tokenizers
 from faster_whisper.audio import decode_audio
 from faster_whisper.feature_extractor import FeatureExtractor
 from faster_whisper.tokenizer import _LANGUAGE_CODES, Tokenizer
-from faster_whisper.utils import download_model, format_timestamp, get_logger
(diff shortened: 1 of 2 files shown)


makaveli10:
@trungkienbkhn looks like this slows down the inference. Have you noticed the increase in latency?

Purfview:
> looks like this slows down the inference

Original author mentioned:

> ...since this also requires extra processing time, we only do this when a probable hallucination is detected.

trungkienbkhn:
@makaveli10 , hello, sorry for the late reply.
I tested an mp3 audio file in 192s with model tiny and device cuda.
If using `hallucination_silence_threshold=2`, total time executing average is 5.38s.
And if don't using this feature, total time is 4.87s.
My code:
```python
model = WhisperModel(model_path, device='cuda')
segments, info = model.transcribe(audio_path, word_timestamps=True, hallucination_silence_threshold=2)
```
=> Latency has increased a bit. But I found that the transcription quality also improved. So I think it's a trade off, it's not too impactful and is acceptable.

PR #667, opened 2024-02-02
update: Ensure method `_asdict` also converts Word object into a native dict

klebster2:
Currently, after segments are generated with the `WhiperModel.transcribe()` method **and** `word_timestamps=True`, upon using method `Segment()._asdict`, we get output like the following:

example_of_word_issue

In the image above, the upper red box is undesired, because `_asdict` should convert to native types. Additionally, this causes information, such as the dictionary keys are also lost when we convert the object that is supposedly a `dict` using `json.dumps` as exemplified in the lower red box.

The change presented therefore ensures that all `Word()` objects are first converted to native types, so that the keys for each `Word()` are retained when the `_asdict` method is called, or `json.dumps(segment._asdict())` is called.

example_of_word_fix

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -5,7 +5,17 @@ import os
 import zlib
 
 from inspect import signature
-from typing import BinaryIO, Iterable, List, NamedTuple, Optional, Tuple, Union
+from typing import (
+    Any,
+    BinaryIO,
+    Dict,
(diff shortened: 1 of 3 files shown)


mgoldey:
This is a nice change that would benefit the community

MahmoudAshraf97:
closed in favor of #1105

Issue #678, opened 2024-02-11
installing requirement pyav==10.* give an error


JiewenGuan:
os:windows11
python: 3.12.1 
attempting to install faster-whisper generates the following error.
` Using cached av-10.0.0.tar.gz (2.4 MB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error

  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [74 lines of output]
      Compiling av\buffer.pyx because it changed.
      [1/1] Cythonizing av\buffer.pyx
      Compiling av\bytesource.pyx because it changed.
      [1/1] Cythonizing av\bytesource.pyx
      Compiling av\descriptor.pyx because it changed.
      [1/1] Cythonizing av\descriptor.pyx
      Compiling av\dictionary.pyx because it changed.
      [1/1] Cythonizing av\dictionary.pyx
      Compiling av\enum.pyx because it changed.
      [1/1] Cythonizing av\enum.pyx
      Compiling av\error.pyx because it changed.
      [1/1] Cythonizing av\error.pyx
      Compiling av\format.pyx because it changed.
      [1/1] Cythonizing av\format.pyx
      Compiling av\frame.pyx because it changed.
      [1/1] Cythonizing av\frame.pyx
      performance hint: av\logging.pyx:232:5: Exception check on 'log_callback' will always require the GIL to be acquired.
      Possible solutions:
          1. Declare 'log_callback' as 'noexcept' if you control the definition and you're sure you don't want the function to raise exceptions.
          2. Use an 'int' return type on 'log_callback' to allow an error code to be returned.

      Error compiling Cython file:
      ------------------------------------------------------------
      ...
      cdef const char *log_context_name(void *ptr) nogil:
          cdef log_context *obj = <log_context*>ptr
          return obj.name

      cdef lib.AVClass log_class
      log_class.item_name = log_context_name
                            ^
      ------------------------------------------------------------

      av\logging.pyx:216:22: Cannot assign type 'const char *(void *) except? NULL nogil' to 'const char *(*)(void *) noexcept nogil'. Exception values are incompatible. Suggest adding 'noexcept' to type 'const char *(void *) except? NULL nogil'.

      Error compiling Cython file:
      ------------------------------------------------------------
      ...

      # Start the magic!
      # We allow the user to fully disable the logging system as it will not play
[...]

Issue #677, opened 2024-02-10
README.md L70 specifies Python 3.8 or greater


swiftraccoon:
https://github.com/SYSTRAN/faster-whisper/blob/f144e4c83d54f3c3304b6a75a3f563e5f84de6cf/README.md?plain=1#L70

I'm unable to install `faster-whisper` on Fedora 39 with python3.12 though. I have to downgrade to python3.9 to be able to install the `faster-whisper` pip package.

```
$ pip install faster-whisper
Defaulting to user installation because normal site-packages is not writeable
Collecting faster-whisper
  Using cached faster-whisper-0.10.0.tar.gz (1.5 MB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting av==10.* (from faster-whisper)
  Using cached av-10.0.0.tar.gz (2.4 MB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [76 lines of output]
      Compiling av/_core.pyx because it changed.
      [1/1] Cythonizing av/_core.pyx
      Compiling av/buffer.pyx because it changed.
      [1/1] Cythonizing av/buffer.pyx
      Compiling av/bytesource.pyx because it changed.
      [1/1] Cythonizing av/bytesource.pyx
      Compiling av/descriptor.pyx because it changed.
      [1/1] Cythonizing av/descriptor.pyx
      Compiling av/dictionary.pyx because it changed.
      [1/1] Cythonizing av/dictionary.pyx
      Compiling av/enum.pyx because it changed.
      [1/1] Cythonizing av/enum.pyx
      Compiling av/error.pyx because it changed.
      [1/1] Cythonizing av/error.pyx
      Compiling av/format.pyx because it changed.
      [1/1] Cythonizing av/format.pyx
      Compiling av/frame.pyx because it changed.
      [1/1] Cythonizing av/frame.pyx
      performance hint: av/logging.pyx:232:5: Exception check on 'log_callback' will always require the GIL to be acquired.
      Possible solutions:
          1. Declare 'log_callback' as 'noexcept' if you control the definition and you're sure you don't want the function to raise exceptions.
          2. Use an 'int' return type on 'log_callback' to allow an error code to be returned.
      
      Error compiling Cython file:
      ------------------------------------------------------------
[...]

Issue #571, opened 2023-11-21
cant install at the moment


tobiasmeyhoefer:
faces this error while installing any help?

pip install faster-whisper                                                                                                                  ✔  16:07:05  
Collecting faster-whisper
  Downloading faster_whisper-0.9.0-py3-none-any.whl.metadata (11 kB)
Collecting av==10.* (from faster-whisper)
  Downloading av-10.0.0.tar.gz (2.4 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 2.4/2.4 MB 9.1 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting ctranslate2<4,>=3.17 (from faster-whisper)
  Downloading ctranslate2-3.21.0-cp312-cp312-macosx_11_0_arm64.whl.metadata (10 kB)
Requirement already satisfied: huggingface-hub>=0.13 in /Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages (from faster-whisper) (0.19.4)
Collecting tokenizers<0.15,>=0.13 (from faster-whisper)
  Downloading tokenizers-0.14.1-cp312-cp312-macosx_11_0_arm64.whl.metadata (6.7 kB)
INFO: pip is looking at multiple versions of faster-whisper to determine which version is compatible with other requirements. This could take a while.
Collecting faster-whisper
  Downloading faster_whisper-0.8.0-py3-none-any.whl.metadata (10 kB)
Collecting tokenizers==0.13.* (from faster-whisper)
  Downloading tokenizers-0.13.3.tar.gz (314 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 314.9/314.9 kB 2.0 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting faster-whisper
  Downloading faster_whisper-0.7.1-py3-none-any.whl.metadata (9.9 kB)
  Downloading faster_whisper-0.7.0-py3-none-any.whl.metadata (9.9 kB)
  Downloading faster_whisper-0.6.0-py3-none-any.whl.metadata (9.1 kB)
  Downloading faster_whisper-0.5.1-py3-none-any.whl (1.5 MB)
[...]

PR #679, opened 2024-02-11
Bumps up PyAV version to support Python 3.12.x

Purfview:
Fix for https://github.com/SYSTRAN/faster-whisper/issues/677 https://github.com/SYSTRAN/faster-whisper/issues/678 https://github.com/SYSTRAN/faster-whisper/issues/571

Bumps up PyAV version to support Python 3.12.x

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-av==10.*
+av==11.*
 ctranslate2>=3.22,<4
 huggingface_hub>=0.13
 tokenizers>=0.13,<0.16


JiewenGuan:
I just tried this fix and it worked, had to install other dependencies manually and install faster-whisper with --no-deps. and for the pytorch I had to install it with 'pip install torchaudio==2.2.0+cu121 --index-url https://download.pytorch.org/whl/cu121' and it runs ok, the torch cuda 118 version also works fine, but after switching torch version the model cache need to be cleared.

Purfview:
> I just tried this fix and it worked, had to install other dependencies manually and install faster-whisper with --no-deps. and for the pytorch I had to install it with 'pip install torchaudio...

But faster-whisper or PyAV doesn't have such dependency. Why you had to install torchaudio?

PR #696, opened 2024-02-21
Bump version to 1.0.0

trungkienbkhn:


BBC-Esq:
Is there going to be a pypi.org updated release later today so I can just pip install faster-whisper 1.0?

PR #706, opened 2024-02-23
Fix window end heuristic for hallucination_silence_threshold

Purfview:
Removes the wishful heuristic which is causing more issues than it's fixing.

Same as https://github.com/openai/whisper/pull/2043

Example of the issue: https://github.com/openai/whisper/pull/1838#issuecomment-1960041500


nguyendc-systran:
Thanks @Purfview for the fix and info

PR #732, opened 2024-03-04
Improve language detection

trungkienbkhn:
Same with https://github.com/openai/whisper/pull/676

PR #768, opened 2024-03-31
Bugfix: code breaks if audio is empty

Purfview:
Regression since https://github.com/SYSTRAN/faster-whisper/pull/732 PR [Improve language detection]

Bug can happen if VAD didn't found voice in audio.
Bugfix brings back the previous behaviour on empty audio.

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -388,8 +388,12 @@ class WhisperModel:
                     language = max(
                         detected_language_info,
                         key=lambda lang: len(detected_language_info[lang]),
+                        default='en',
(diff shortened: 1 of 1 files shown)


trungkienbkhn:
@Purfview , thanks for your findings. However, I think that it can revert to its previous behavior with a simple logic outlined below. (`seek <= content_frames`):
```python
while (
    seek <= content_frames
    and seek
    < self.feature_extractor.nb_max_frames * language_detection_segments
)
```

Purfview:
Yes, 'seek <= content_frames' is simpler fix, thx.

PR #769, opened 2024-03-31
Foolproof: Disable VAD if clip_timestamps is in use

Purfview:
Prevent silly things to happen.

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -280,6 +280,7 @@ class WhisperModel:
           clip_timestamps: Union[str, List[float]]
             Comma-separated list start,end,start,end,... timestamps (in seconds) of clips to
              process. The last end timestamp defaults to the end of the file.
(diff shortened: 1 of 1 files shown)


trungkienbkhn:
@Purfview , could you provide an example of the strange behavior if using both vad_filter and clip_timestamp ?

Purfview:
> @Purfview , could you provide an example of the strange behavior if using both vad_filter and clip_timestamp ?

`clip_timestamps` is like custom/manual VAD and it runs after Silero VAD code. Them both running at same time doesn't make sense.

For example you want to transcribe only the second minute of audio and set`clip_timestamps="120,180"`, if VAD is enabled then audio can be different and the second minute wouldn't be at "120,180" anymore., transcription would happen at some unknown portion of audio.

trungkienbkhn:
Okay. But I think you should add a comment or logging for this update logic code: vad_filter will be ignored if clip_timestamp is used.

Purfview:
Comment added.

PR #773, opened 2024-04-01
Benchmarks

trungkienbkhn:
# Benchmarks
This PR introduces functionality to measure benchmarking for memory, Word Error Rate (WER), and speed in Faster-whisper.  
### GPU
Use Python thread with the [py3nvml](https://github.com/fbcotter/py3nvml) module to monitor GPU memory. This approach will continuously track GPU memory consumption at set intervals, then return the maximum memory used for the inference function. 
### RAM
Use [memory_profiler](https://github.com/pythonprofilers/memory_profiler) module to measure maximum increase of memory usage.
## 2. WER
Evaluate the Faster-whisper model on the [LibriSpeech](https://huggingface.co/datasets/librispeech_asr) validation-clean dataset with [streaming mode](https://huggingface.co/blog/audio-datasets#streaming-mode-the-silver-bullet), meaning no audio data has to be downloaded to your local device.
Use [jiwer](https://github.com/jitsi/jiwer) and [evaluate](https://github.com/huggingface/evaluate) modules from HF to calculate WER.
## 3. Speed
Calculate the min of `args.repeat x time-averaged` over 10 inference function calls.

**_Feel free to edit/update more benchmarking methods !_**


BBC-Esq:
I'd like to suggest using the ```psutil``` library directly instead of ```memory_profiler```, which inherently uses ```psutil``` under the hood and hasn't been updated in years.  Also, I'd suggest using the official "pynvml" library, which is actually named ```nvidia-ml-py```; it's more up-to-date.

https://pypi.org/project/nvidia-ml-py/

Here's an example of how I'm using it:

https://github.com/BBC-Esq/VectorDB-Plugin-for-LM-Studio/blob/main/src/metrics_bar.py

trungkienbkhn:
@BBC-Esq , tks for your suggestion. But I think your proposed pynvml module seems not official, it is built from this [repo](https://github.com/wookayin/nvidia-ml-py). I tried both (py3nvml + nvidia-ml-py) and they gave the same results.
I used [py3nvml](https://github.com/huggingface/transformers/blob/5080ab12c818d3875858ad37b667c00c6f09f094/src/transformers/benchmark/benchmark.py#L44) like in transformer repo from HF.

BBC-Esq:
This screenshot says that it's maintained by nvidia so I'm not sure what the confusion is...
image

"py3nvml" on pypi says that some individual maintains it...

image

If it works, it works...but when I went to the repo for py3nvml the last time it was updated was awhile ago...

Any feedback on ```psutil``` versus ```memory_profiler```.  All the same perhaps?

BBC-Esq:
As an aside, it's my understanding that "nvidia-ml-py" (i.e. the one I'm using) imports as "pynvml" NOT nvidia-ml-py, which is what confused me initially.  Multiple forks exist, some that have created pypi packages named similarly so...I'm about 85%-95% sure I have the official one.  ;-)

BBC-Esq:
Don't know if it's helpful, but here's a script of mine that collects more than just GPU VRAM usage...might be helpful for benchmarking.  I've found that gpu/cuda usage is a useful metric as well as power usage...

https://github.com/BBC-Esq/Nvidia_Gpu_Monitor/blob/main/metrics_pynvml.py

BBC-Esq:
If Ctranslate2 ever supports RocM, that'd be great, and you want to benchmark on AMD gpus I've struggled unsuccessfully to do it.  I don't own an AMD gpu, almost bought one solely to program, but anyways here's where my research left off...

https://github.com/BBC-Esq/VectorDB-Plugin-for-LM-Studio/issues/130

trungkienbkhn:
> This screenshot says that it's maintained by nvidia so I'm not sure what the confusion is...

`NVIDIA Corporation` can be setup in the [author](https://github.com/wookayin/nvidia-ml-py/blob/main/setup.py#L53) field of the setup.py file.
> Don't know if it's helpful, but here's a script of mine that collects more than just GPU VRAM usage...might be helpful for benchmarking

Tks, I will look at this.

PR #804, opened 2024-04-24
Loosen tokenizers version constraint

Keating950:
Constraining the version of tokenizers to `<0.16` makes faster-whisper incompatible with the latest version of transformers, and the tests all pass when using the latest version (`0.19.1`).


bzikst:
I do confirm the problem, please consider merging this PR, please.

PR #816, opened 2024-05-04
Bump version to 1.0.2

trungkienbkhn:


ddorian:
If possible to also include https://github.com/SYSTRAN/faster-whisper/pull/819 in the new version.

PR #817, opened 2024-05-04
Clarify documentation for hotwords

Keating950:
I think this phrasing is clearer. Also, if you're open to a change in behaviour, I would consider throwing an exception if mutually incompatible options are provided (e.g. `hotwords` and `prefix`). I can add a commit doing this if you're interested.

I've added some other minor doc cleanups as well.


Keating950:
I've added some other minor doc cleanups.

PR #867, opened 2024-06-04
Improve language detection when using clip_timestamps

ben91lin:
Using clip_timestamps to improve the initial seek for language detection, avoiding incorrect detection at the start of the audio file.

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -370,16 +370,27 @@ class WhisperModel:
                     or language_detection_segments < 1
                 ):
                     language_detection_segments = 1
-                seek = 0
-                detected_language_info = {}
+                start_timestamp = (
+                    float(clip_timestamps.split(",")[0])
+                    if isinstance(clip_timestamps, str)
+                    else clip_timestamps[0]
+                )
                 content_frames = (
                     features.shape[-1] - self.feature_extractor.nb_max_frames
                 )
-                while (
-                    seek <= content_frames
-                    and seek
-                    < self.feature_extractor.nb_max_frames * language_detection_segments
-                ):
+                seek = (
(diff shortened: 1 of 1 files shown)


faster_whisper/transcribe.py
```diff
@@ -370,16 +370,24 @@ def transcribe(
                     or language_detection_segments < 1
                 ):
                     language_detection_segments = 1
-                seek = 0
+                if isinstance(clip_timestamps, str):
+                    start_timestamp = float(clip_timestamps.split(",")[0])
+                else:
+                    start_timestamp = clip_timestamps[0]
                 detected_language_info = {}
+                seek = int(
```
trungkienbkhn, line 378: Should replace with:
```python3
seek = int(start_timestamp * self.frames_per_second)
```

ben91lin:
```
# If seek is beyond all frames, set it to the last segment.
if seek >= features.shape[-1]:
    seek = content_frames
```

If audio_length is 80s and start_timestamp is 67s, it will clip the last 1.3 seconds for detection. 
If start_timestamp greater or equal 80s, force use the last nb_max_frames for detection.

trungkienbkhn:
> ```
> # If seek is beyond all frames, set it to the last segment.
> if seek >= features.shape[-1]:
>     seek = content_frames
> ```

I think it's wrong logic. In this case, if `seek = content_frame`, then fw will detect language from `content_frames` to `features.shape[-1]`.
=> These are padded values (meaningless), and can lead to incorrect language detection.
Should set `seek = 0` if `seek >= content_frames`

ben91lin:
I think you are right, I omitted the padding from FeatureExtractor. Setting `seek = 0` if `seek >= content_frame` is the correct approach.

PR #898, opened 2024-07-03
Filter out non_speech_tokens in suppressed tokens

jordimas:
This matches the implementation of OpenIA Whisper suppress_tokens with faster-whisper.

In the default behaviour (-1) it filters out the non_speech_tokens which should help to filter out non-relevant text to the transcription.

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -1233,7 +1233,6 @@ def get_suppressed_tokens(
     tokenizer: Tokenizer,
     suppress_tokens: Optional[List[int]],
 ) -> Optional[List[int]]:
-
     suppress_tokens = list(suppress_tokens)
 
     if -1 in suppress_tokens:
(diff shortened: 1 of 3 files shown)


faster_whisper/transcribe.py
```diff
@@ -1229,13 +1233,16 @@ def get_suppressed_tokens(
     tokenizer: Tokenizer,
     suppress_tokens: Optional[List[int]],
 ) -> Optional[List[int]]:
```
trungkienbkhn, line 1235: -> Tuple[int]

faster_whisper/transcribe.py
```diff
@@ -1229,13 +1233,16 @@ def get_suppressed_tokens(
...
-    if not suppress_tokens or -1 in suppress_tokens:
-        return suppress_tokens
-
     suppress_tokens = list(suppress_tokens)
```
trungkienbkhn, line 1236: Remove unnecessary logic

faster_whisper/transcribe.py
jordimas: Fixes, thanks

faster_whisper/transcribe.py
jordimas: Fixes, thanks

Issue #919, opened 2024-07-20
Segment timestamps are buggy in BatchedInferencePipeline


kalradivyanshu:
In the recent code #856, `BatchedInferencePipeline` the segments are sometimes wrong for some reason.

I recorded an audio clip: https://drive.google.com/file/d/1cbDbiXi12SIsd0hIDfs61VdtgI78Fg_p/view?usp=sharing

and for this `BatchedInferencePipeline` gives `Segment(id=1, seek=2307, start=18.71, end=23.07...` even though the clip is only 3 second long.

Here is a colab recreating this issue, just upload batch.wav from the link above: https://colab.research.google.com/drive/1ie7uMFW_LJUvxGHW3KkT5iG8uUZiTVwU?usp=sharing

```py
segments, info = batched_model.transcribe(arr, word_timestamps=True, batch_size = 1)
print(list(segments))
```
output:
```
[Segment(id=1, seek=2307, start=18.71, end=23.07, text=' Hey Michael, how are you?', tokens=[14690, 3899, 11, 703, 389, 345, 30], avg_logprob=-0.21215820871293545, compression_ratio=0.7647058823529411, no_speech_prob=0.08087158203125, words=[Word(start=18.71, end=20.11, word=' Hey', probability=0.71240234375), Word(start=20.11, end=20.11, word=' Michael,', probability=0.732421875), Word(start=22.71, end=22.71, word=' how', probability=0.98046875), Word(start=22.71, end=22.71, word=' are', probability=0.9990234375), Word(start=22.71, end=23.07, word=' you?', probability=0.99853515625)], temperature=1.0)]
```

```py
segments, info = model.transcribe(arr, word_timestamps=True)
print(list(segments))
```
correct output:
```
[...]

PR #920, opened 2024-07-21
fix word timestamps for batched inference

MahmoudAshraf97:
it was caused by wrong `num_frames` argument when finding the alignments, it was assumed that inferring it from encoder output size was sufficient but turned out to cause issues such as #919 when the actual segment size is much less that the inferred size


MahmoudAshraf97:
closed in favour of #921

PR #921, opened 2024-07-21
Remove the usage of `transformers.pipeline` from `BatchedInferencePipeline` and fix word timestamps for batched inference

### LINKED ISSUE #919: Segment timestamps are buggy in BatchedInferencePipeline

MahmoudAshraf97:
This PR removes Remove the usage of `transformers.pipeline` from `BatchedInferencePipeline` because there's no need to use it in the first place
this simplifies the code and removes a requirement
it also fixes #919 
it was caused by wrong `num_frames` argument when finding the alignments, it was assumed that inferring it from encoder output size was sufficient but turned out to cause issues such as #919 when the actual segment size is much less that the inferred size

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -211,9 +211,12 @@ class BatchedInferencePipeline(Pipeline):
             model_inputs["features"], self.tokenizer, forward_params
         )
 
-        segment_size = encoder_output.shape[1] * 2
         segmented_outputs = []
+        segment_sizes = []
         for segment_metadata, output in zip(model_inputs["seg_metadata"], outputs):
+            duration = segment_metadata["end_time"] - segment_metadata["start_time"]
+            segment_size = int(duration * self.model.frames_per_second)
+            segment_sizes.append(segment_size)
             (
                 subsegments,
                 seek,
@@ -223,8 +226,7 @@ class BatchedInferencePipeline(Pipeline):
                 tokens=output["tokens"],
                 time_offset=segment_metadata["start_time"],
                 segment_size=segment_size,
-                segment_duration=segment_metadata["end_time"]
-                - segment_metadata["start_time"],
+                segment_duration=duration,
                 seek=0,
             )
             segmented_outputs.append(
@@ -248,7 +250,7 @@ class BatchedInferencePipeline(Pipeline):
                 segmented_outputs,
                 self.tokenizer,
                 encoder_output,
-                segment_size,
+                segment_sizes,
                 forward_params["prepend_punctuations"],
                 forward_params["append_punctuations"],
                 self.last_speech_timestamp,
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -15,8 +15,7 @@ import tokenizers
 import torch
 
 from pyannote.audio import Model
-from transformers import Pipeline
-from transformers.pipelines.pt_utils import PipelineIterator
+from tqdm import tqdm
 
 from faster_whisper.audio import decode_audio, pad_or_trim
 from faster_whisper.feature_extractor import FeatureExtractor
@@ -105,7 +104,7 @@ class TranscriptionInfo(NamedTuple):
 # (https://github.com/m-bain/whisperX) and adapted for faster_whisper
 
 
-class BatchedInferencePipeline(Pipeline):
+class BatchedInferencePipeline:
     """
     Huggingface Pipeline wrapper for WhisperModel.
     Copyright (c) 2022, Max Bain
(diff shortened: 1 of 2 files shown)


faster_whisper/transcribe.py
```diff
@@ -621,10 +509,27 @@ def transcribe(
             all_language_probs=all_language_probs,
         )
 
+        audio_segments, segments_metadata = self.audio_split(
+            audio, vad_segments, sampling_rate
```
Jiltseb, line 513: Can you make the check for shorter audio as the first condition check after `if not_vad_segments ` in line 439? This is to avoid VAD error such as https://github.com/m-bain/whisperX/issues/844 by first checking the condition: 
```
if duration < self.chunk_length:
                vad_segments = [
                    {"start": 0.0, "end": duration, "segments": [(0.0, duration)]}
                ]
```

Jiltseb (changes requested):
I have tested the newer version for speed. Slighly slower on youtube-commons subset but has similar results for an internal benchmarking dataset. Have a look at the comment on VAD issue.

faster_whisper/transcribe.py
MahmoudAshraf97: I think the problem originates from the VAD itself, doing this will solve it for short audio files, but what about long audio files where this issue occurs in a later window? I'm waiting for the user who reported the problem to upload the test file so I can confirm this

Jiltseb: The issue is that VAD is not able to detect it as a voiced region (considering it a noisy region). We can enable `vad_offset` and `vad_onset` user configurable, to aid it. The audio file, for example, is: https://litter.catbox.moe/kyu2q8.wav

MahmoudAshraf97: the file is 404, i've asked the user to reupload it
I'm going to see if silero will detect it correctly, and then decide how to move forward

MahmoudAshraf97: as I expected, the problem is in the vad model
this is the sequential output
`Hey, Cardage, do you know Candace? If you don't know who I am, tell me! BOOM! I GOT DANGER!`
this is the batched output using silero vad instead of pyannote
`Cardage, do you know Candace?`
batched version using pyannote finds no speech in the audio

If you listen carefully to the audio you'll find that only the first sentence makes sense while the rest is very noisy that the model is just hallucinating
this is the ground truth:
`Hey Carnage, do you know Candice? YOUR MOTHER HUNG HERSELF, I got that adrenaline momentum`

Jiltseb: We could address it in detail in the Silero VAD PR. For the time being, we can avoid VAD computation for shorter files with duration < `self.chunk_length`.

MahmoudAshraf97: the VAD refactor PR is almost ready, so I suggest we merge this and address the issue in the next PR if it's still a concern

MahmoudAshraf97:
@trungkienbkhn can I get a final review and merge if no changes needed?

trungkienbkhn:
@MahmoudAshraf97 , LGTM. Tks for your contribution.

faster_whisper/transcribe.py
trungkienbkhn: To enhance flexibility, I think we could add a new arg to allow to specify the VAD model (pyannote/silero). This arg will be applicable in both sequential and multi-batch modes.
There have also been some recent feedback that silero v5 is worse than v4 in some languages. So we could also enable to choose the specific version of the VAD model by directly specifying its path.

Issue #934, opened 2024-07-26
1.0.3 VAD v5 is much worse than 1.0.2 VAD v4 in korean


zx3777:
Large portions of the speech are missing.

Some files have subtitles files of 10kb using version 1.0.2, while only less than 1kb using version 1.0.3.

This video file
https://www.youtube.com/watch?v=tVLOBfzbJV8
resulted in 320 lines of subtitles using version 1.0.2, but only 218 lines using version 1.0.3. Many conversations were not recognized in version 1.0.3.

I only compared Korean, other languages ​​have not been tested yet.

zx3777:
This is the audio file for the video above.
https://mega.nz/file/QacS2LCJ#x_Gq9GgV8aPk2qRVskfzNBuyM9XAI-Pv2SBIwxfomnk

x86Gr:
I agree, I also have worse performance, just not as much, however the overall WER for non english speech is going down. Go back to silero or at least let us choose the VAD model

zx3777:
> I agree, I also have worse performance, just not as much, however the . Go back to silero or at least let us choose the VAD model

Version 1.0.3 release still uses silero, but with an upgraded version.
WER going down maybe because the VAD only identifies sufficiently clear speech.

MahmoudAshraf97:
@zx3777 that will cause higher WER, a missing word is still an error to count
You should try playing with the vad settings and see how it makes a difference, the model was changed but the parameters are still tuned for the previous one

zx3777:
> @zx3777 that will cause higher WER, a missing word is still an error to count You should try playing with the vad settings and see how it makes a difference, the model was changed but the parameters are still tuned for the previous one

I tried `--vad_threshold 0.4 0.3 0.2` in 1.0.3, and there was a slight improvement, but the recognized subtitles are still much less than in 1.0.2.

hoonlight:
Hi, could you try again on the master branch and let me know the results?

x86Gr:
I will run the tests on our audio corporas, with different parameters, but it won't be quick

zx3777:
> Hi, could you try again with the master branch and let me know the results?

I tested the master branch version before the upgrade to [New PR for Faster Whisper: Batching Support, Speed Boosts, and Quality Enhancements], and the results were the same.

In my opinion, after the new PR, only the batched version uses a different VAD implementation. The normal version still uses the VAD from 1.03, so the results should be the same.

hoonlight:
[...]

Issue #925, opened 2024-07-22
vadOptions default value


ngcheeyuan:
Just putting it out here. It's a good idea to tune your VadOptions if you are facing some issues with it.

Working with non-english inputs,

The default value for VadOptions is a little to aggressive, and I personally recommend a longer padding and lower threshold values.

For padding, what I found was that it needed a longer padding because it tends to cut off the first few syllabus of the words if it was too aggressive.

ooobo:
Be nice to hear what values have worked well for other users (for all languages) - maybe a discussion would be better?

ckgithub2019:
> Just putting it out here. It's a good idea to tune your VadOptions if you are facing some issues with it.
> 
> Working with non-english inputs,
> 
> The default value for VadOptions is a little too aggressive, and I personally recommend a longer padding and lower threshold values.
> 
> For padding, what I found was that it needed a longer padding because it tends to cut off the first few syllabus of the words if it was too aggressive.

a longer padding means a longer "speech_pad_ms"? By default, the speech_pad_ms is 400, should be changed to 800 or even higher? what's your suggestion?

ngcheeyuan:
> longer padding means a longer "speech_pad_ms"? By default, the speech_pad_ms is 400, should be changed to 800 or even higher?

Yes. I went with 1000ms and 0.4 threshold iirc. Do listen to the audio before and after silerovad is applied.

PR #936, opened 2024-07-28
Use Silero VAD in Batched Mode

MahmoudAshraf97:
This PR tries to close the gap between Batched and Sequential versions
Summary of Changes:
* Reimplementation of Silero model and inference is now 3x faster
* Batched pipeline now uses silero instead of pyannote vad, this reduces the amount of code in the repo that is needed to handle two vad models
* added a script to evaluate WER on Youtube Commons ASR dataset (useful for longform and batched evaluation)
* Unified the batched and sequential `transcribe` function as much as I could

WER Comparisons

Batched (`without_timestamps=True`, `vad_filter=True`, `chunk_length=25`) on Youtube Commons using `distil-large-v3`:
Before: WER: 13.910
After: WER: 13.712

Vad Parameters are not completely tuned, but I don't have the resources to evaluate on multilingual datasets, contributions are welcome


hoonlight:
When I used the batch version, I got better transcribe results than the sequential version, but I'm not sure if this is due to pyannote VAD or if there is an additional process in the batch version to improve WER. I was wondering if you have ever compared Silero VAD v3 with pyannote VAD.

By the way, thank you for your contribution to improve faster-whisper. Even if it's a well-discussed and approved PR, someone is entitled to complain, but no one has the right to rudely point fingers.

Jiltseb:
> When I used the batch version, I got better transcription results compared to the sequential version. I'm not sure if this is due to pyannote VAD or if there is an additional process in the batch version that improves WER. Have you ever compared Silero VAD with pyannote VAD?
> 
> By the way, thank you for your contribution to improving faster-whisper. Even though it was a well-discussed and approved PR, anyone is entitled to have their opinion about it, but no one has the right to be rude.

It is indeed possible to have better results for long-form transcription in batched mode. This is because there is no context passing between batches and it prevents ambiguous text from the previous context passing to the next set of frames for computation.

Thanks for your kind words regarding the batched PR.

@MahmoudAshraf97 I would suggest adding the numbers with pyannote VAD and silero VAD (WER and the speed-up) for completeness.

zh-plus:
Is it better to let users choose the VAD model from pyannote VAD or Silero VAD?

I get better VAD segments for Chinese & Japanese audios with pyannote than Silero, even though I try hard to tune the VAD-related parameters for Silero.

Other users have also encountered this kinda issue: #925, #934.

Jiltseb:
> Is it better to let users choose the VAD model from pyannote VAD or Silero VAD?
> 
> I get better VAD segments for Chinese & Japanese audios with pyannote than Silero, even though I try hard to tune the VAD-related parameters for Silero.
> 
> Other users have also encountered this kinda issue: #925, #934.

Pyannote model could be superior VAD, but the extra dependency on pyannote and torch is a concern at the moment.

MahmoudAshraf97:
@zh-plus it can be an option of course, but keeping pyannote will force us to keep pytorch in the requirements which we are trying to remove based on users feedback, i'm trying to think of a structure to make the whole batching thing optional with optional dependencies for those who want it

MahmoudAshraf97:
Performance numbers added, tests are passing locally but are failing on CI because torchaudio can't find a backend to use since they are not installed after the removal of `pyannote` (along with 78 other packages so I guess it's a win)
this PR should not be merged until we do one of the following:
1. Include `soundfile` or `sox` in the requirements as a backend to `torchaudio`
2. Revert back to using `PyAV` and use manual garbage collection to avoid the resampler memory leak if needed, this will make us one step closer to removing torch completely

ozancaglayan:
Thanks for the PR!

Could you add your script that exports the Silero V5 model to encoder and decoder ONNX files?

faster_whisper/vad.py
```diff
@@ -35,8 +28,9 @@ class VadOptions(NamedTuple):
       speech_pad_ms: Final speech chunks are padded by speech_pad_ms each side
     """
 
-    threshold: float = 0.5
-    min_speech_duration_ms: int = 250
+    onset: float = 0.5
+    offset: float = onset - 0.15
+    min_speech_duration_ms: int = 0
```
ozancaglayan, line 33: Can you maybe leave these options (threshold, onset, offset) as they were, e.g. not rename them as it would break signature & parameter passing APIs?

Why are you changing min_speech_duration_ms to 0? I think 250ms is a sane default otherwise you may end up with segments that are very small for having speech inside, maybe even empty ones?

MahmoudAshraf97: it's best to give the users the freedom to tune the parameters as they wish, previously `offset` was fixed to `threshold - 0.15`, but now users have the option to tune it as they with without having to play with the code internals, it might not be backwards compatible but it's a very minimal change to adapt

as for `min_speech_duration_ms`, benchmarks (YT Commons and Librispeech) showed that dropping it from 250 to 0 had minimal positive or no effect on sequential inference, but it had a very positive impact on batched inference as it combines segments differently than the sequential

ozancaglayan:
Thanks. Out of curiosity did you find those reference implementations elsewhere or did you rewrite them based on JIT'ted or is there a way to automatically generate from JIT'ted models?

MahmoudAshraf97:
> Thanks. Out of curiosity did you find those reference implementations elsewhere or did you rewrite them based on JIT'ted or is there a way to automatically generate from JIT'ted models?
> 
> PS: OK I think you can get the compiled graph from `.code` variables but that one does not seem to be a pure python implementation.

I reimplemented it from scratch based on what I could understand from the JITed code and mapped the weights manually using the dictionary, both implementations are within 1e-5 tolerance from the original implementation

MahmoudAshraf97:
> Performance numbers added, tests are passing locally but are failing on CI because torchaudio can't find a backend to use since they are not installed after the removal of `pyannote` (along with 78 other packages so I guess it's a win) this PR should not be merged until we do one of the following:
> 
> 1. Include `soundfile` or `sox` in the requirements as a backend to `torchaudio`
> 2. Revert back to using `PyAV` and use manual garbage collection to avoid the resampler memory leak if needed, this will make us one step closer to removing torch completely

Reverted back to PyAV in #961, once it is merged then this one is merged we can get rid of torch dependency

Jiltseb:
> > Performance numbers added, tests are passing locally but are failing on CI because torchaudio can't find a backend to use since they are not installed after the removal of `pyannote` (along with 78 other packages so I guess it's a win) this PR should not be merged until we do one of the following:
> > 
> > 1. Include `soundfile` or `sox` in the requirements as a backend to `torchaudio`
> > 2. Revert back to using `PyAV` and use manual garbage collection to avoid the resampler memory leak if needed, this will make us one step closer to removing torch completely
> 
> Reverted back to PyAV in #961, once it is merged then this one is merged we can get rid of torch dependency

Nice. I have also re-implemented numpy version to get rid of torch dependency. But will stick to this for getting rid of the torch in two steps. I will test the memory leakage and report in https://github.com/SYSTRAN/faster-whisper/pull/961.

kenho211:
Encounter another error for audio without speech. Not the same one as in https://github.com/SYSTRAN/faster-whisper/pull/973
` File "/home/ubuntu/.local/lib/python3.10/site-packages/faster_whisper/transcribe.py", line 362, in transcribe
    clip_timestamps = merge_segments(active_segments, vad_parameters)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/faster_whisper/vad.py", line 315, in merge_segments
    curr_start = segments_list[0]["start"]
IndexError: list index out of range`

Can we just return an empty list in `merge_segments` if `segment_list` is empty?

MahmoudAshraf97:
should be fixed now

hobodrifterdavid:
Hi. I'm running a lot of audio through the batch transcribe function, getting a couple of exceptions on some files:

image

image

Appreciate the work guys.

MahmoudAshraf97:
@hobodrifterdavid can you upload audios that reproduce the two exceptions?

hobodrifterdavid:
I don't. I just added a check to make sure the audio clips I am sending are at least 5s long (it's possible I was requesting transcription of some zero-length files), and I will improve the logging to record what is processing when an error occurs, will let you know if I see the error again.

carolinaxxxxx:
@MahmoudAshraf97 good job sir. However I have a problem - it does not respect the --hotwords option at all in standard interference and batching mode. I tested on different materials. In the "old" version before the batching mode was introduced, --hotwords worked very well. I'm using your forks now.

zazpie:
@MahmoudAshraf97 any timeline for when this changes will be merged? We are currently pinning `faster-whisper` dependency to `master` since there hasn't been a new version.

The latest changes did increased our speed significantly hence why we are waiting for those changes to be merged and stabilized.

Also, any way we could help?

MahmoudAshraf97:
This PR is currently pending @trungkienbkhn response and review, along with #961

carolinaxxxxx:
@MahmoudAshraf97 any chance to make hotwords work as it did before? Thanks. Good job.

MahmoudAshraf97:
Hi @carolinaxxxxx , please open a separate issue with before and after comparison, also upload an audio file and steps to reproduce so we can identify the issue.
All previous PRs did not change anything related to hotwords so this is unrelated to this PR

eschmidbauer:
im also experience this issue- seems like an urgent fix. Can we get it merged?

toanhn:
Can we use this branch? I see it still "Open".

danghieuan:
Thank for your contributed. Hope this PR approved.

toanhn:
I hope any update can improve the accuracy of transcription.

faster_whisper/transcribe.py
```diff
@@ -227,10 +225,7 @@ def audio_split(audio, segments, sampling_rate):
 
     def transcribe(
         self,
-        audio: Union[str, torch.Tensor, np.ndarray],
-        vad_segments: Optional[List[dict]] = None,
```
Jiltseb, line 290: Wouldn't it be possible to optionally pass in the `vad_segments` to support any external VAD models as needed? This can default to `None` to make it look as close as a sequential transcribe. This is beneficial if you have multiple transcription requests and the framework can process VAD of next audio while doing transcription of current audio.

MahmoudAshraf97: This is already possible using `clip_timestamps`, why did I remove `vad_segments` and added `clip_timestamps`? I'm trying to make `transcribe` functions in both batched and sequential as close as possible and interchangeable

Jiltseb (commented):
Added some minor comments. I have tested Silero on batched version and got similar WER, but the speed is 60% slower compared to previous VAD.  This is on a test set of 9 youtube videos with various audio types and a length from 3-13 minutes. With Silero, it is still at least 2x faster than sequential version. With pyannote VAD it was 3.8x faster.

Have you seen this speed difference?

faster_whisper/vad.py
```diff
@@ -293,42 +304,13 @@ def __call__(
...
-# and adapted for faster_whisper.
-
-
 def merge_segments(segments_list, vad_options: VadOptions):
     curr_end = 0
     seg_idxs = []
     merged_segments = []
-    edge_padding = vad_options.speech_pad_ms / 1000
-    chunk_length = vad_options.max_speech_duration_s
+    sampling_rate = 16000
```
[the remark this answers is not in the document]
MahmoudAshraf97: if the distance between two segments is less than `2 * edge_padding` they are merged together, so it's guaranteed that no audio is included twice, I found that the increasing or decreasing the padding value didn't make much difference so I left it as is to account for higher error margin

As for the speedups, I found that both implementations to be almost identical or within measuring error range, my specs are:
i7 12700k
RTX 3070 Ti
32GB Ram

Although even if silero implementation is slightly slower, it's worth it because of the simpler requirements and the increased code reuse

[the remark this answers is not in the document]
Jiltseb: Makes sense for the `edge_padding` and agree that Silero makes the codebase lean and easy to maintain. Do you have the audio file you tested?

Jiltseb (commented):
Add sampling_rate as an argument in `merge_segments` function as well and remove hard coded sampling rate (L318)

faster_whisper/vad.py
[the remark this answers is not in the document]
MahmoudAshraf97: I test on the yt commons dataset
pyannote vad:
Evaluating...: 94it [25:32, 16.31s/it]
WER: 13.976

Silero Vad:
Evaluating...: 94it [26:22, 16.83s/it]
WER: 13.756

toanhuynhnguyen:
I get the below error when updating this PR:
Unable to load any of {libcudnn_ops.so.9.1.0, libcudnn_ops.so.9.1, libcudnn_ops.so.9, libcudnn_ops.so}
Invalid handle. Cannot load symbol cudnnCreateTensorDescriptor
Aborted (core dumped)

toanhuynhnguyen:
@MahmoudAshraf97 How can I install this update on the master branch, can the below command line work:
pip3 install --force-reinstall "faster-whisper @ https://github.com/SYSTRAN/faster-whisper/archive/refs/heads/master.tar.gz"

toanhuynhnguyen:
After installing:
```
pip install --force-reinstall "faster-whisper @ https://github.com/SYSTRAN/faster-whisper/archive/refs/heads/master.tar.gz"
```

I run the code:
```
from faster_whisper import WhisperModel, BatchedInferencePipeline

model = WhisperModel("medium", device="cuda", compute_type="float16")
batched_model = BatchedInferencePipeline(model=model)
segments, info = batched_model.transcribe("audio.mp3", batch_size=16)

for segment in segments:
    print("[%.2fs -> %.2fs] %s" % (segment.start, segment.end, segment.text))
```

I get this error:
```
Unable to load any of {libcudnn_ops.so.9.1.0, libcudnn_ops.so.9.1, libcudnn_ops.so.9, libcudnn_ops.so}
Invalid handle. Cannot load symbol cudnnCreateTensorDescriptor
Aborted (core dumped)
```

Can anyone help me with this error, thank you so much.

PR #856, opened 2024-05-24
New PR for Faster Whisper: Batching Support, Speed Boosts, and Quality Enhancements

Jiltseb:
Hello everyone,

This PR adds a major update to Faster Whisper, bringing both speed and quality improvements!

### Speed improvements:

- **Batching support**: Inspired by [whisper-x](https://github.com/m-bain/whisperX), this update introduces batching support allowing for a 3x speed increase. This implementation builds on whiper-x and supports more run-time arguments and external VAD segments. The batched version now runs at 64x real-time speed, compared to the previous 20x.

- **Faster feature extraction**: We've incorporated torchaudio-based parallel STFT as an alternative to the current implementation from [transformers](https://github.com/huggingface/transformers/blob/main/src/transformers/models/whisper/feature_extraction_whisper.py), providing additional speed boosts. With the `enable_ta_fe` flag, the final version achieves an impressive 104x real-time speed. This is up to 12.5x on average compared to OpenAI implementation!

Using the batched version is straightforward:
```python
from faster_whisper import WhisperModel, BatchedInferencePipeline
#load faster-whisper model in the usual way
model = WhisperModel("medium", device="cuda", compute_type="float16") 

#apply batched pipeline
batched_model = BatchedInferencePipeline(model=model)

#predict using the batched_model
result = batched_model.transcribe("audio.mp3", batch_size=16)

for segment, info in result:
	print("[%.2fs -> %.2fs] %s" % (segment.start, segment.end, segment.text))
```										

### Quality Improvements

1.  **Consistency across runs**: By setting the model seed, consistency across runs is improved.
2. **Reducing hallucinations**: Stricter checks in the inference pipeline reduce unstructured or repeated phrases.
3. **Reliable language detection**: A new function detects language more reliably by considering highly confident and random segments, breaking ties to determine the major language.
4. **Code-switching support**: Handles audio with multiple languages by detecting language every 30 seconds and dynamically directing data flow. Since the exact language switching position is unknown, this can have an error within a 30 sec segment range.

Language detection Usage:
```python
from faster_whisper import WhisperModel

model = WhisperModel("medium", device="cuda", compute_type="float16")
language_info = model.detect_language_multi_segment("audio.mp3")
```
#### A. Open source benchmarking:

[...]

PR #961, opened 2024-08-14
revert back to using PyAV instead of torch audio

MahmoudAshraf97:
This PR reverts the torchaudio code that was added in #856 and removes `torchaudio` dependency but still keeps `torch`

the reason why `torch` wasn't removed in this PR is that feature extraction still depends on it and I didn't want to include the numpy feature extraction in this PR to keep it simple and to reduce the number of conflicts to be resolved with #936 

this should be merged before #936, after both are merged, a new PR will be created to completely remove the `torch` dependency

### DIFF
--- a/faster_whisper/audio.py
+++ b/faster_whisper/audio.py
@@ -1,7 +1,20 @@
+"""We use the PyAV library to decode the audio: https://github.com/PyAV-Org/PyAV
+
+The advantage of PyAV is that it bundles the FFmpeg libraries so there is no additional
+system dependencies. FFmpeg does not need to be installed on the system.
+
+However, the API is quite low-level so we need to manipulate audio frames directly.
+"""
+
+import gc
+import io
+import itertools
+
 from typing import BinaryIO, Union
 
+import av
+import numpy as np
 import torch
-import torchaudio
 
 
 def decode_audio(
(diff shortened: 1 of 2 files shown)


joiemoie:
Will removing torch remove the supposed FFT speedup?

MahmoudAshraf97:
> Will removing torch remove the supposed FFT speedup?

Yes, but I have a new implementation using numpy in progress that is faster than the original implementation 
These are the performance figures on my machine for a 30s segment:
Current Torch on cpu vs new numpy vs old numpy
image_20240817_022356948a5b27-82c6-46cd-a932-266a7b0610f8

I'm still investigating the possibility of GPU acceleration

joiemoie:
Great to hear! I work with segments about 10 seconds long, so no benefit from batching. However, I am curious and possibly interested into bumping up to the latest commit because of this FFT speedup

MahmoudAshraf97:
these are the latest performance figures for the new FE @joiemoie 
```
# torch on CPU
4.7 ms ± 179 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
# torch on GPU
689 µs ± 9.41 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
# New numpy on CPU
30.7 ms ± 1.4 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
# New CuPY on GPU
5.31 ms ± 138 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
# Old numpy on CPU
61.1 ms ± 2.86 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

ozancaglayan:
Do you take into account the overhead of moving long audio tensors to GPU when doing the above measurements on GPU?

ozancaglayan:
Another option could actually be to move the feature extraction layer to CTranslate2 as well, whisper.cpp has some implementations
https://github.com/ggerganov/whisper.cpp/blob/master/src/whisper-mel-cuda.cu

MahmoudAshraf97:
> Another option could actually be to move the feature extraction layer to CTranslate2 as well, whisper.cpp has some implementations https://github.com/ggerganov/whisper.cpp/blob/master/src/whisper-mel-cuda.cu

https://github.com/OpenNMT/CTranslate2/issues/1419#issuecomment-1682321352

carolinaxxxxx:
@MahmoudAshraf97 good job sir. However I have a problem - it does not respect the --hotwords option at all in standard interference and batching mode. I tested on different materials. In the "old" version before the batching mode was introduced, --hotwords worked very well.

Asmar097 (approved):
Good job

PR #973, opened 2024-08-22
Return empty generator when no active speech in transcribe

kenho211:
To fix
`/lib/python3.10/site-packages/faster_whisper/transcribe.py", line 523, in transcribe features = torch.stack( RuntimeError: stack expects a non-empty TensorList`

when there is no active speech, i.e. `vad_segments` after `merge_chunks` is empty.


hobodrifterdavid:
I was getting errors thrown when processing some audio files with the latest commit, reverted back, now it's okay:

image

I think that was the error, I guess this is the fix?

MahmoudAshraf97:
it was fixed already in #936

PR #980, opened 2024-08-29
fix:The timestamps for the second and subsequent segments are incorrect

caiwuu:
When VAD detects multiple segments of speech in an audio clip, the timestamps from the second segment onward are incorrect, as shown in the image below
This is incorrect：
image
This is correct：
image

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -2085,9 +2085,11 @@ def restore_speech_timestamps(
             )
 
         else:
+            middle = (segment.start + segment.end) / 2
+            chunk_index = ts_map.get_chunk_index(middle)
             segment = segment._replace(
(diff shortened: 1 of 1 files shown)


faster_whisper/transcribe.py
```diff
@@ -2085,9 +2085,11 @@ def restore_speech_timestamps(
             )
 
         else:
+            middle = (segment.start + segment.end) / 2
+            chunk_index = ts_map.get_chunk_index(middle)
```
MahmoudAshraf97, line 2089: this is not necessarily correct, as you noticed it fails the tests as it produces wrong timestamps in the test file, you need to identify the reason for the wrong timestamps in the first place to decide whether the incorrect timestamps are caused by the pre/post processing or from the whisper model

PR #1014, opened 2024-09-16
Update transcribe.py

hedrergudene:
provide an early exit in case no voice activity is detected.


MahmoudAshraf97:
already solved in #936

PR #1038, opened 2024-10-05
Update feature_extractor.py

BBC-Esq:
Added mel filter bank caching to FeatureExtractor class to optimize memory usage and reduce computational overhead when processing multiple audio files with identical parameters, particularly beneficial for batch processing scenarios.


faster_whisper/feature_extractor.py
```diff
@@ -54,61 +77,88 @@ def get_mel_filters(sr, n_fft, n_mels=128):
...
-        )
+        # Move waveform to the target device if necessary
+        if self.device == "cuda" and not waveform.is_cuda:
+            waveform = waveform.to(self.device)
 
+        # Apply padding if required
         if padding:
-            waveform = torch.nn.functional.pad(waveform, (0, self.n_samples))
-
-        window = torch.hann_window(self.n_fft).to(waveform.device)
```
abodacs, line 99: Why was the `hann_window` deleted?

faster_whisper/feature_extractor.py
```diff
@@ -54,61 +77,88 @@ def get_mel_filters(sr, n_fft, n_mels=128):
...
         if waveform.dtype is not torch.float32:
             waveform = waveform.to(torch.float32)
 
-        waveform = (
-            waveform.to(self.device)
-            if self.device == "cuda" and not waveform.is_cuda
-            else waveform
-        )
+        # Move waveform to the target device if necessary
+        if self.device == "cuda" and not waveform.is_cuda:
```
abodacs, line 136: Improved readability, thank you.

abodacs (changes requested):
@BBC-Esq 
Thanks for your work

faster_whisper/feature_extractor.py
BBC-Esq: I'll take a look...

BBC-Esq: This is actually an optimization. In the old version, a new Hann window was being created and moved to the device every time __call__ was executed. The new version creates it once during initialization and caches it as an instance variable (self.window).

MahmoudAshraf97:
Hello, the way you implemented caching is not going to work because the only parameter that might change when creating is `n_mels`, to change this parameter you need to initialize a new `FeatureExtractor` instance which will clear cache thus invalidating its purpose. Anyways, creating window takes around 6us which represents less than 0.05% of the execution time and the mel filters is created only once thus there is no need for caching.

In the feature, it's preferable to use a caching decorator such as `functools.lru_cache` as an easy caching solution instead of implementing it as a dictionary

PR #1039, opened 2024-10-05
Update vad.py

BBC-Esq:
Pre-convert the audio tensor to a NumPy array and pad it upfront. This change eliminates the need to conditionally pad each audio chunk within the loop, thereby reducing computational overhead and speeding up the processing, especially for long audio files.


MahmoudAshraf97:
Hi and thanks for the contribution, this code was entirely removed in 936, so the fix is no more relevant

PR #1053, opened 2024-10-10
Adding the num of segments into TranscriptionInfo

stsfaroz:
Users will find it useful to know the number of segments.

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -98,6 +98,7 @@ class TranscriptionInfo(NamedTuple):
     all_language_probs: Optional[List[Tuple[str, float]]]
     transcription_options: TranscriptionOptions
     vad_options: VadOptions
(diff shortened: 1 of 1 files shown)


stsfaroz:
@MahmoudAshraf97 can you look into this failing in run test , I'm not sure of it

MahmoudAshraf97:
This will not work, segments is a generator whose length is unknown until it finishes, so `len(segments)` will run the whole transcription just to get the number of segments and then return an empty generator hence the error that is appearing in the CI because the transcription is empty,

Anyway, I don't see why this should be added to the codebase when you can easily do it later after getting the transcription results

stsfaroz:
@MahmoudAshraf97 I want to do it in cpu , but this len(list(segments)) takes more seconds than the model.transcribe , is there a way to get it faster ?

PR #1056, opened 2024-10-11
Add language_detection_only parameter

Frenzie:
Depending a bit on the length of the audio and the number of language_detection_segments this can save quite a bit of time. In my testing a minimum of about 20% at 3-4 s, going up to 80% at 10 s.


MahmoudAshraf97:
you can already do that by using `model.detect_language(audio)`

Frenzie:
Thank you. I feel silly I hadn't noticed because I was working with the Pip version. It looks like there have been some pretty big changes.

PR #1082, opened 2024-10-23
Update requirements.txt: latest release from ctranslate2 from 10-22-2024 breaks faster-whisper

kristopher-smith:
latest release from ctranslate2 from 10-22-2024 breaks faster-whisper. See release history here: https://pypi.org/project/ctranslate2/4.4.0/#history

I have edited the requirements.txt file to remedy this and limited the range from 4.0, 4.4.0 which is the latest working version for ctranslate2 with faster-whisper instead of 4.0, 5

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-ctranslate2>=4.0,<5
+ctranslate2>=4.0,<=4.4.0
 huggingface_hub>=0.13
 tokenizers>=0.13,<1
 onnxruntime>=1.14,<2 


MahmoudAshraf97:
faster whisper can use ctranslate2==4.5.0 so there is no reason to bound the version, the problem is caused by a mismatched CuDNN version caused by pytorch
either use ct2<4.5 along with torch<2.4 or ct2==4.5 along with torch>=2.4

MahmoudAshraf97:
I might've been quick to close this PR, the problem is more complicated than I expected, `v4.5.0` works just fine with pip installed cudnn, but if you have a torch version where the cuda binaries are precompiled such as `torch==2.5.0+cu121` or any version that ends with `+cu12`, this error comes up, the only solution is downgrade to v`4.4.0` at the moment which is strange because it was compiled using cudnn 8.9

BBC-Esq:
Here is the compatibility issue...resolving dependency issues sucks gonads...but here's my attempt:

### Torch Versions and Supported CUDA

| **Torch Version** | **Supported CUDA Versions** |
|-------------------|-----------------------------|
| 2.5.0             | 11.8, 12.1, 12.4            |
| 2.4.1             | 11.8, 12.1, 12.4            |
| 2.4.0             | 11.8, 12.1, 12.4            |
| 2.3.1             | 11.8, 12.1                  |


* NOTE: It's my understanding that ```torch``` does NOT have wheels for CUDA 12.3, for example.  When they publish wheels that have in their name something like "cu123" it refers to compatibility with CUDA 12.3 only...not CUDA 12.3 or higher...

### cuDNN Versions and Supported CUDA

| **cuDNN Version** | **Supported CUDA Versions** |
|-------------------|-----------------------------|
| 8.9.7             | 11 through 12.2             |
| 9.0.0             | 11 through 12.3             |
| 9.1.0             | 11 through 12.4             |
| 9.2.0             | 11 through 12.5             |
| 9.2.1             | 11 through 12.5             |
| 9.3.0             | 11 through 12.6             |
| 9.4.0             | 11 through 12.6             |
| 9.5.0             | 11 through 12.6             |


* NOTE:  This information comes from the "compatibility matrix" on Nvidia's website.

Based on the foregoing...
1) ```Ctranslate2''' 4.5.0 or higher requires cuDNN 9+
2) cuDNN 9+ supports as low as ```torch``` 12.3.
3) However, ```torch``` does NOT have a wheel that supports torch 12.3...only 12.1 and 12.4.

BBC-Esq:
To follow up...

The same holds true fur ```ctranslate2``` AND ANY OTHER LIBRARIES YOU USE IN YOUR PROGRAM.

For example, if ```ctranslate2``` installed a ```torch``` version that's incompatible with other dependencies, you'd need to use the ```--no-deps``` flag when using ```pip``` and then individually install its dependencies separately.  It gets quite annoying...

MahmoudAshraf97:
As this is not going to be solvable, workarounds are in #1086 , Thanks @BBC-Esq and @kristopher-smith

kristopher-smith:
Thanks for looking into this so promptly everybody.

MahmoudAshraf97:
The reason we can't do that is that the problem is not from faster whisper, it's an incompatibility between ctranslate2 and user specific cuda installation, so if we limit it to `4.4.0` all users that use torch 2.4.0 or higher will not be able to install it correctly

kristopher-smith:
> The reason we can't do that is that the problem is not from faster whisper, it's an incompatibility between ctranslate2 and user specific cuda installation, so if we limit it to `4.4.0` all users that use torch 2.4.0 or higher will not be able to install it correctly

Good point. This could be handled in the setup.py file maybe?

I am happy to take a crack at this with a pull request which handles your library compatibility matrix during install.

PR #1091, opened 2024-10-25
Update audio.py

BBC-Esq:
Utilize NumPy's optimized array concatenation (np.concatenate), which is faster and more memory-efficient.

### DIFF
--- a/faster_whisper/audio.py
+++ b/faster_whisper/audio.py
@@ -41,7 +41,7 @@ def decode_audio(
         rate=sampling_rate,
     )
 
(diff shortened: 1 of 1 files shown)


MahmoudAshraf97:
Hi, 
The CI tests are failing due to formatting errors and functional errors
you can follow the [Contributing Guideline](https://github.com/SYSTRAN/faster-whisper/blob/master/CONTRIBUTING.md) to run tests locally and make sure they pass.

Also can you provide simple benchmark numbers before and after this change, maybe including loading time and memory consumption for several audio lengths

PR #1094, opened 2024-10-26
Add modal integration for STT model

amanjain97:


Jiltseb:
Looks like you have opened PR on original repo instead of your fork master!

PR #1096, opened 2024-10-27
[DRAFT] refactor of transcribe.py[DISCUSS]

BBC-Esq:
This is an initial DRAFT to float/discuss the idea of using a centralized configuration class.  It reduces the current code base of transcribe.py by approximately 12,000 character while, in theory, achieving the same functionality.  I'm referring to ~12,000 character reduction in code logic, not comments or docstrings.  Consolidating some of those as well should further reduce the code base.

Again, a jumping off point for a discussion. Thanks!

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -9,6 +9,7 @@ from collections import Counter, defaultdict
 from inspect import signature
 from math import ceil
 from typing import BinaryIO, Iterable, List, NamedTuple, Optional, Tuple, Union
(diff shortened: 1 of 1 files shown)


MahmoudAshraf97:
I suggest we turn this into an issue or a discussion where we can share ideas, this is much more understandable than a PR

BBC-Esq:
okay, not sure how to do that...Can you convert it or...

MahmoudAshraf97:
Close this and open an issue or a discussion

Issue #171, opened 2023-04-23
Is it possible to add audio context length parameter like in whisper.cpp


fquirin:
I'm working mostly with very small files like 2-6s and what I noticed many times in Whisper is that transcription time is not linear. What I mean is that 3s and 6s files will be processed in almost the same time. I don't know the inner parts of Whisper well enough but I think this is related to the 30s window used to process chunks.

In whisper.cpp there is a parameter called `audio_ctx` that has been described here: https://github.com/ggerganov/whisper.cpp/issues/137

Do you think it might be possible to implement something similar in faster-whisper so that we can experiment with it and see if it reduces processing time for small files?

guillaumekln:
The Whisper model requires a fixed 30-second window as input. It was trained like that so we have to do the same during inference. For now I don't plan on making this size configurable as it would probably degrade the transcription accuracy.

guillaumekln:
> If I remember correctly for smaller files the rest is filled up with zeros.

The audio is filled with zero, but the log-Mel spectrogram and ultimately the encoder outputs are non-zero for these padding positions.

Maybe you can experiment with whisper.cpp and report here the impact on speed and accuracy (e.g. WER)?

fquirin:
Unfortunately whisper.cpp has the audio_ctx only for streaming mode, so I can't test it on my reference WAV files, but here are some insights and experiences I could gain:
- Georgi himself said that ac `768` (ms) is probably the lower limit before accuracy starts to degrade considerably.
- In streaming mode, 768 is a mix of good results and lost words for me, probably because the context transfer from one window to the next isn't happening properly or data simply gets lost. For example counting from 1 to 10 will likely result in something like "1, 2, 3, 4, 6, 7, 8, 9". Some users claim that they don't have the same issue but I could never verify this.
- In the whisper.cpp community the 'ac' parameter is considered to increase speed considerably, but it is hard to test without being able to use it on WAV files and get some reproducible numbers. To me it "feels" faster as well.

Another quote by Georgi: `I reduced the audio context by half and the performance doubled. jfk.wav transcribed correctly!`

juanmc2005:
Hello, I'm facing a similar issue. I'm willing to sacrifice a bit of accuracy for faster inference for streaming but this is currently not possible to configure.

[...]

PR #1101, opened 2024-10-28
Use correct features padding for encoder input

MahmoudAshraf97:
This adheres faster whisper implementation to the original OpenAI implementation as discussed in #1084 

These are WER comparisons before and after
These figures can be reproduced by running `benchmarks/yt_commons.py` and switching the batched inference to sequential
```python
word_timestamps=False,
without_timestamps=True,
vad_filter=True,
```

| Model               | Before WER | After WER  |
|---------------------|------------|------------|
| distil-large-v3     | 26.277     | **14.762** |
| distil-large-v2     | **71.848** | 81.456    |
| distil-medium.en    | 68.044 | **66.565**     |
| distil-small.en     | 68.719     | **67.220** |

the performance regression of `distil-large-v2` can be ignored because `distil-large-v3` should be used as a drop-in replacement

This should also affect all use cases where chunk length is less than 30 for all models
There is also an average WER improvement by around 2% (relative) for batched inference across all models

| Model               | Before WER | After WER  |
|---------------------|------------|------------|
| tiny.en             | 15.437 | 15.063     |
| tiny                | 21.765 | 21.390     |
| base.en             | 14.300 | 13.816     |
| base                | 17.709 | 17.251     |
| small.en            | 13.054     | 12.617 |
| small               | 16.413 | 16.088    |
| medium.en           | 13.299     | 12.894 |
| medium              | 15.991 | 15.593     |
| large-v1            | 19.458     | 19.590 |
| large-v2            | 15.237     | 15.148 |
| large-v3            | 16.514     | 15.997 |
| large-v3-turbo      | 14.576     | 14.044 |
| distil-small.en     | 14.004     | 13.918 |
| distil-medium.en    | 14.074 | 13.972     |
| distil-large-v2     | 13.419 | 13.574     |
| distil-large-v3     | 13.688     | 13.533 |

### DIFF
--- a/faster_whisper/audio.py
+++ b/faster_whisper/audio.py
@@ -109,9 +109,9 @@ def _resample_frames(frames, resampler):
         yield from resampler.resample(frame)
 
 
-def pad_or_trim(array, length: int, *, axis: int = -1):
+def pad_or_trim(array, length: int = 3000, *, axis: int = -1):
     """
-    Pad or trim the audio array to N_SAMPLES, as expected by the encoder.
+    Pad or trim the Mel features array to 3000, as expected by the encoder.
     """
     axis = axis % array.ndim
     if array.shape[axis] > length:
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -441,9 +441,11 @@ class BatchedInferencePipeline:
         features = (
             torch.stack(
                 [
-                    self.model.feature_extractor(chunk, to_cpu=to_cpu)[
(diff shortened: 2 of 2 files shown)


MahmoudAshraf97:
This PR disabled the ability to change the encoder input and output dimensions (`audio_ctx`), any `chunk_length` passed to the `transcribe` will be respected, but the encoder input will be padded to 30s equivalent regardless of the chunk length

https://github.com/SYSTRAN/faster-whisper/issues/171#issuecomment-1701177072

PR #1104, opened 2024-10-30
Add Pydantic typing

extrange:
Hi, we are using this library extensively at our company, and internally we have added type validation with Pydantic. If this is useful, I can update the documentation as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ __pycache__/
 # Distribution / Packaging
 venv/
 *.egg-info
+build/
 
 # Unit Test
 .pytest_cache/
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ __pycache__/
 
 # Distribution / Packaging
 venv/
+*.egg-info
 
 # Unit Test
 .pytest_cache/
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -8,13 +8,24 @@ import zlib
 from collections import Counter, defaultdict
 from inspect import signature
 from math import ceil
-from typing import BinaryIO, Iterable, List, NamedTuple, Optional, Tuple, Union
+from typing import (
+    BinaryIO,
+    Iterable,
+    List,
+    Literal,
+    NamedTuple,
+    Optional,
+    Sequence,
+    Tuple,
+    Union,
+)
 
 import ctranslate2
 import numpy as np
 import tokenizers
 import torch
 
+from pydantic import BaseModel
 from tqdm import tqdm
 
 from faster_whisper.audio import decode_audio, pad_or_trim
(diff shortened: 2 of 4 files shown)


MahmoudAshraf97:
Hello, I'm not well experienced with pydantic tbh, can you explain how this is useful or helps ease future maintainability? I know there is a lot of code duplication that this reduces but is there other benefits?

extrange:
We have a use case where we need to serialize the output `tuple[Iterable[Segment], TranscriptionInfo]` into JSON, and then deserialize it back into a typed Python object/model.

Namedtuples do not natively allow nested deserialization. For example, using the following JSON:

```python
segments, info = model.transcribe(audio, word_timestamps=True)
out = [s_asdict() for s in segments]
print(json.dumps(out, indent=2))
```

```json5
[
  {
    "id": 1,
    "seek": 2490,
    "start": 0.0,
    "end": 6.0,
    "text": " Ladies and gentlemen, thank you for being here and for your written representations.",
    "tokens": [
      50364,
      //...
    ],
    "avg_logprob": -0.26516544380608725,
    "compression_ratio": 1.597883597883598,
    "no_speech_prob": 0.01361083984375,
    "words": null,
    "temperature": 0.0
  },
//...
]

```
If we try to convert this back to a `list[Segment]`, the `words` is not converted:

```python
from faster_whisper.transcribe import Segment
[Segment(**s) for s in parsed]
```
```python
Segment(
    id=1,
    seek=2490,
    start=0.0,
    end=4.62,
    text=" Ladies and gentlemen, thank you for being here and for your written representations.",
    tokens=[
        50364,
# ...
    ],
    avg_logprob=-0.26516544380608725,
    compression_ratio=1.597883597883598,
    no_speech_prob=0.01361083984375,
    words=[
        [0.0, 0.32, " Ladies", 0.310546875],
        [0.32, 0.54, " and", 0.89453125],
# ...
    ],
    temperature=0.0,
)

```

Currently, we are [recursing through the `Segment` and `TranscriptionInfo` objects](https://github.com/extrange/transcription-benchmarks/blob/705a097a493f6c32086fa434b8f326cfee099e8b/src/transcription_benchmarks/inference/systran_faster_whisper_large_v2/inference.py#L333-L357) to convert them to Pydantic models (which natively allow nested deserialization), but it would be nice if the library supported this out of the box.

MahmoudAshraf97:
I see, but why not use `dataclass`? it supports recursive serialization too and will not add an extra dependency or change much in the code

extrange:
I think dataclasses are not serializable to JSON, at least not without [another library](https://github.com/lidatong/dataclasses-json):

```python
import json
from dataclasses import dataclass


@dataclass
class Foo:
    x: str


foo = Foo(x="bar")
json.dumps(foo)

# TypeError: Object of type Foo is not JSON serializable
```

extrange:
Closed in favour of #1105

PR #1105, opened 2024-10-30
replace `NamedTuple` with `dataclass`

MahmoudAshraf97:
this is to allow conversion of `Segment` and `Words` to `dict` or `json` without recursively going through them
Similar to #667 and #1104 

Unlike NamedTuples, dataclasses are not iterable, so they need to be converted to a dict first

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -6,9 +6,10 @@ import random
 import zlib
 
 from collections import Counter, defaultdict
+from dataclasses import asdict, dataclass
 from inspect import signature
 from math import ceil
-from typing import BinaryIO, Iterable, List, NamedTuple, Optional, Tuple, Union
+from typing import BinaryIO, Iterable, List, Optional, Tuple, Union
 
 import ctranslate2
 import numpy as np
@@ -30,14 +31,19 @@ from faster_whisper.vad import (
 )
 
 
-class Word(NamedTuple):
+@dataclass
+class Word:
     start: float
     end: float
     word: str
     probability: float
 
+    def _asdict(self):
+        return asdict(self)
 
-class Segment(NamedTuple):
+
+@dataclass
+class Segment:
     id: int
     seek: int
     start: float
@@ -50,9 +56,13 @@ class Segment(NamedTuple):
     words: Optional[List[Word]]
     temperature: Optional[float] = 1.0
 
+    def _asdict(self):
+        return asdict(self)
+
 
 # Added additional parameters for multilingual videos and fixes below
-class TranscriptionOptions(NamedTuple):
+@dataclass
+class TranscriptionOptions:
     beam_size: int
     best_of: int
     patience: float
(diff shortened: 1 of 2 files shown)


MahmoudAshraf97:
@extrange

extrange:
Thanks for looking at this!

I'm not too familiar with dataclasses, but it looks like there's an [issue](https://stackoverflow.com/a/69059343) with parsing nested dataclasses, for example:

```python
import json
from dataclasses import dataclass, asdict

@dataclass
class Bar:
    y: str

@dataclass
class Foo:
    x: str
    bar: Bar

baz = Foo(x="x", bar=Bar(y="y"))

out = json.dumps(asdict(baz))
# '{"x": "x", "bar": {"y": "y"}}'
```

The parameter `bar` is not parsed, nor type-checked:

```python
Foo(**json.loads(out))
# Foo(x='x', bar={'y': 'y'})
```

We can do it manually, but this will break if new keys are added:

```python
obj = json.loads(out)
Foo(x=obj["x"], bar=Bar(**obj["bar"]))
# Foo(x='x', bar=Bar(y='y'))
```

MahmoudAshraf97:
I didn't consider the parsing case tbh, is there a use case in your mind that needs converting the json back to a dataclass in faster whisper? as far as I'm concerned, these classes are pretty stable and no keys are being added or removed since a long time ago so the manual solution might not be so bad after all

This PR solves the serialization problem with minimal code changes and no additional dependencies, Pydantic might be the best solution, but I think it's overkill in this repo and requires a lot of changes that I might not be comfortable with

extrange:
The main use case for us is when working with AWS SageMaker endpoints, for which we have to serialize parameters and deserialize the output of the transcription.

For now we're using a [helper library](https://github.com/nuhs-projects/faster-whisper-types) which we've pinned to specific commits of faster-whisper.

MahmoudAshraf97:
You can still use `Pydantic` validation, at least for type checking
```python
import pydantic
from faster_whisper.transcribe import Segment

@pydantic.dataclasses.dataclass
class Segment(Segment):
    ...

Segment(
    **{
        "id": "invalid_id",
        "seek": 0,
        "start": 0.0,
        "end": 1.0,
        "text": "transcription",
        "tokens": [1, 2, 3],
        "avg_logprob": 0.1,
        "compression_ratio": 0.1,
        "no_speech_prob": 0.1,
        "words": None,
        "temperature": 1.0,
    }
)
```
```
ValidationError: 1 validation error for Segment
id
  Input should be a valid integer, unable to parse string as an integer [type=int_parsing, input_value='invalid_id', input_type=str]
    For further information visit https://errors.pydantic.dev/2.9/v/int_parsing
```

When this PR is merged, you can use the wrapper to add pydantic type validation to the existing dataclass instances, these instances can be passed to any function that expects an instance of the original classes

extrange:
That's a nice solution. Pydantic supports [nested stdlib dataclasses](https://docs.pydantic.dev/latest/concepts/dataclasses/#inherit-from-stdlib-dataclasses) so I think this will also work for us (since maybe the deserialization use case isn't sufficient to have pydantic as an additional dependency).

PR #1141, opened 2024-11-14
fix: Use correct `seek` value in output, fix word timestamps when the initial timestamp is not zero

### LINKED ISSUE #1140: Some segment has a 1 second shifted after PR #856
appreciate your hard work

---

audio (2 minutes): [01.aac.zip](https://github.com/user-attachments/files/17751633/01.aac.zip)

The correct SRT result (using commit fbcf58b, which is before the huge PR #856): [01.old.srt.zip](https://github.com/user-attachments/files/17751733/01.old.srt.zip)

The wrong SRT result (using latest commit 85e61ea): [01.new.srt.zip](https://github.com/user-attachments/files/17751755/01.new.srt.zip)

---

I am **not** using the batch version

```python
model = faster_whisper.WhisperModel(
    model_size_or_path='large-v2',
    device='cuda',
    cpu_threads=4,
)
model.transcribe(
    audio=audio,
    language=None,
    task='transcribe',
    vad_filter=False,
    initial_prompt=None,
    word_timestamps=True,
    repetition_penalty=1.0,
)
```

script from this project https://github.com/heimoshuiyu/whisper-fastapi

---

image

some segments on the left (wrong) has 1 second mismatch (shift +1s) than the right (correct)

---

I also test on the commit of RP #856 (eb839023), which is worse

result SRT: 
[01.eb839023.srt.zip](https://github.com/user-attachments/files/17752205/01.eb839023.srt.zip)

image

left: commit eb839023 PR #856
middle: latest commit 85e61ea
right: commit fbcf58b

heimoshuiyu:
this fix #1140 

code from commit before #856 <https://github.com/SYSTRAN/faster-whisper/blob/fbcf58bf984a309edab9d21e23bca872e1d5fe61/faster_whisper/transcribe.py#L1051-L1055>

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -174,7 +174,9 @@ class BatchedInferencePipeline:
                         compression_ratio=get_compression_ratio(
                             self.tokenizer.decode(subsegment["tokens"])
                         ),
-                        seek=seek,
+                        seek=int(
+                            chunk_metadata["start_time"] * self.model.frames_per_second
+                        ),
                     )
                     for subsegment in subsegments
                 ]
@@ -497,7 +499,7 @@ class BatchedInferencePipeline:
                 for segment in result:
                     seg_idx += 1
                     yield Segment(
-                        seek=int(result[-1]["end"] * self.model.frames_per_second),
+                        seek=segment["seek"],
                         id=seg_idx,
                         text=segment["text"],
                         start=round(segment["start"], 3),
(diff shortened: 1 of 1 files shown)


MahmoudAshraf97:
Hi and thanks for your efforts, however in the example audio that you uploaded, `seek / hop_length * sampling_rate` and `segment[0]["start"]` are both equal to `5998` which is expected, so the error must be caused by something else, it's strange that this fixed it because as I said earlier, the two values should be interchangeable.

I'll continue investigating this

MahmoudAshraf97 (approved):
Thanks for the PR. I've made some additional fixes to address the core issue.

Purfview:
Finally faster-whisper is fixed after that #856 mess. Good job!

PR #1181, opened 2024-11-29
Brings back original VAD parameters naming

Purfview:
For no reason in v1.1.0 VAD parameter `threshold` was changed to `onset`  and `neg_threshold` to `offset`. 

This PR brings original naming back.


MahmoudAshraf97 (approved):
That is more inline with original silero naming

Purfview:
> That is more inline with original silero naming

What's more important that it's inline with pre-1.1.0 faster-whisper, renaming parameters breaks many projects for no reason.

> As for the failing CI, tokenizers dropped support for python 3.8 in the latest release...

Just switch CI to 3.9 I think, anyway, savvy users will be able solve compatibility issues, I doubt that newbies use 3.8. :)

PR #1183, opened 2024-12-01
constraint tokenizers version for python 3.8

MahmoudAshraf97:
since support for python 3.8 was removed in version 0.21.0 of tokenizers, this constraints it to the latest supported version


MahmoudAshraf97:
Decided to switch CI to 3.9 instead, around 8% of the users use 3.8 as per [This survey](https://lp.jetbrains.com/python-developers-survey-2023/#python-versions) so it's not worth it to keep compatibility

PR #1186, opened 2024-12-03
Add duration of audio and VAD removed duration to BatchedInferencePipeline

greenw0lf:
With the non-batched version of the WhisperModel, you would get logging output like:

```
Processing audio with duration 01:33:59.990
VAD filter removed 06:55.648 of audio
```

Whereas, when calling the BatchedInferencePipeline's `transcribe()` method, that is no longer the case.

This PR tries to bring that logging back as I believe it does not add extra overhead to the logging and it is quite useful for developers who wish to know how much audio gets processed in the end.

If this is not the case and it is an issue related to my usage of the model, I apologize in advance!


faster_whisper/transcribe.py
```diff
@@ -114,6 +114,7 @@ def __init__(
         self,
         model,
     ):
+        self.logger = get_logger()
```
MahmoudAshraf97, line 117: `self.model` already has a logger, so I'd rather we use the same logger instead of having duplicate loggers in both the pipeline and the model

MahmoudAshraf97 (changes requested):

MahmoudAshraf97 (approved):

PR #1188, opened 2024-12-04
change language_detection_threshold type to float

extrange:
`language_detection_threshold` can't be supplied as `None` as `detect_language` compares it with a `float`.

Change parameter type in `WhisperModel` to reflect this.

Added test to confirm this behavior.

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -731,7 +731,7 @@ class WhisperModel:
         clip_timestamps: Union[str, List[float]] = "0",
         hallucination_silence_threshold: Optional[float] = None,
         hotwords: Optional[str] = None,
-        language_detection_threshold: Optional[float] = 0.5,
+        language_detection_threshold: float = 0.5,
         language_detection_segments: int = 1,
     ) -> Tuple[Iterable[Segment], TranscriptionInfo]:
         """Transcribes an input file.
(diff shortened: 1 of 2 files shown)


extrange:
There's no use case for `None`, so the `Optional` has been removed. Furthermore, supplying `None` raises a `TypeError` as the test shows. So, we shouldn't allow `None` to be supplied for type safety.

MahmoudAshraf97:
I see your point, thanks for noticing
for consistency can you also review other occurrences of this case? these are some examples
https://github.com/SYSTRAN/faster-whisper/blob/8327d8cc647266ed66f6cd878cf97eccface7351/faster_whisper/transcribe.py#L298

also I don't think the test is needed, the number of invalid input combos is too large to have tests for all of them, what do you think?

kenho211:
In the original suggestion https://github.com/openai/whisper/pull/676, if `language_detection_threshold` is None, "a majority voting on the segment languages in order to decide the language" will be performed.

Would like to add this back because I want to force `detect_language` to do majority vote for a long audio.

kenho211:
> Hello, what is the use case where it needs to be supplied as `None`?

From https://github.com/openai/whisper/pull/676, the use case of `None` should be to conduct a majority vote for however many segments allowed to determine the top language.
For the current implementation, we can use `language_detection_threshold = 1` to do the same.

PR #1191, opened 2024-12-06
Fix neg_threshold

Purfview:
Fix neg_threshold.
Now it should work properly, before its default was hardcoded to 0.35

### DIFF
--- a/faster_whisper/vad.py
+++ b/faster_whisper/vad.py
@@ -35,7 +35,7 @@ class VadOptions:
     """
 
     threshold: float = 0.5
-    neg_threshold: float = threshold - 0.15
+    neg_threshold: float = None
     min_speech_duration_ms: int = 0
     max_speech_duration_s: float = float("inf")
     min_silence_duration_ms: int = 2000
@@ -91,7 +91,8 @@ def get_speech_timestamps(
     triggered = False
     speeches = []
     current_speech = {}
-    neg_threshold = max(neg_threshold, 0.01)
+    if neg_threshold == None:
+        neg_threshold = max(threshold - 0.15, 0.01)
 
     # to save potential segment end (and tolerate some silence)
     temp_end = 0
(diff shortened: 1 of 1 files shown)


MahmoudAshraf97:
the default was `0.35` indeed but it's not hardcoded and can changed when initializing `VadOptions`
a simpler solution would be
```diff
    threshold: float = 0.5
-    neg_threshold: float = threshold - 0.15
+    neg_threshold: float = max(threshold - 0.15, 0)
    min_speech_duration_ms: int = 0
    max_speech_duration_s: float = float("inf")
    min_silence_duration_ms: int = 2000
    speech_pad_ms: int = 400
```
this would remove the need for other changes as it will not be `None` and if it's set explicitly to a negative it will behave as zero, so no need to handle this

Purfview:
> but it's not hardcoded

```
vad_parameters = {}
vad_parameters['threshold'] = 0.2

model.transcribe(fileaudio, vad_filter=True, vad_parameters=vad_parameters)

# prints below this line https://github.com/SYSTRAN/faster-whisper/blob/f32c0e8af3f2843bf70071d3d14a480a6a23da66/faster_whisper/vad.py#L93
# vad.py_print threshold: 0.2
# vad.py_print neg_threshold: 0.35   <- hardcoded, should be 0.05
```

>since there is no issue if it is a negative value, then even the solution above is not needed, it will be merely cosmetic

Yeah, I didn't looked if there is an issue, just that it's that way in original silero -> https://github.com/snakers4/silero-vad/blob/3f9fffc26121fe0bb819370d50443389d0b7c8c6/src/silero_vad/utils_vad.py#L310C1-L311C52

abedkhooli:
I can't set the threshold itself in v 1.1.0 (and can't read it from vad.VadOptions).

Purfview:
> I can't set the threshold itself in v 1.1.0 (and can't read it from vad.VadOptions).

Look there: https://github.com/SYSTRAN/faster-whisper/pull/1181

MahmoudAshraf97 (approved):

PR #1194, opened 2024-12-09
Make batched suppress_tokens behaviour same as in sequential

Purfview:
Make batched suppress_tokens behaviour same as in sequential mode.

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -495,7 +495,11 @@ class BatchedInferencePipeline:
             initial_prompt=initial_prompt,
             prefix=prefix,
             suppress_blank=suppress_blank,
-            suppress_tokens=get_suppressed_tokens(tokenizer, suppress_tokens),
+            suppress_tokens=(
+                get_suppressed_tokens(tokenizer, suppress_tokens)
+                if suppress_tokens
+                else suppress_tokens
+            ),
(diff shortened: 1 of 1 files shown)


MahmoudAshraf97:
They should have the same behavior, the `else` case is already handled in the function itself
https://github.com/SYSTRAN/faster-whisper/blob/8327d8cc647266ed66f6cd878cf97eccface7351/faster_whisper/transcribe.py#L1836-L1837
I'd suggest modifying the sequential code instead for consistency

Purfview:
> They should have the same behavior, the else case is already handled in the function itself

Not same, as the function populates suppress_tokens.
I think original author had in mind an option to have suppress_tokens empty.

Look at https://github.com/SYSTRAN/faster-whisper/pull/898

jordimas:
Hello!
I have very little time, but from what I remember there are 3 uses cases:
- User provides None tokens (nothing is applied)
- User provides -1 and then the defaults are used
- The user provides a list and then the defaults + user's list is used

Consider also that adding suppress is not called is the if evaluates to False  (e.g. None)
https://github.com/openai/whisper/blob/90db0de1896c23cbfaf0c58bc2d30665f709f170/whisper/decoding.py#L557C25-L557C40

What I can tell is that I test extensively OpenIA and faster-whisper to behave the same in all cases when I did the PR.

Issue #1169, opened 2024-11-23
Version 1.1.0 has onnxruntime thread affinity crash


Appfinity-development:
Updated from 1.0.3 to 1.1.0. Now an onnxruntime thread affinity crash occurs each time. Both versions run on a Nvidia A40 with 4 CPU cores, 48GB VRAM and 16GB RAM (on  a private Replicate server). Shouldn't be a hardware issue. Our model config:

```
  self.whisper_model = WhisperModel(
            "large-v2",
            device="cuda",
            compute_type="float16",
            cpu_threads=4,
            num_workers=1
        )
        
        ...
        
         options = dict(
            vad_filter=True,
            vad_parameters=dict(min_silence_duration_ms=1000),
            initial_prompt=prompt,
            word_timestamps=True,
            language=language,
            log_progress=True,
            hotwords=prompt
        )
        
        segments, transcript_info = self.whisper_model.transcribe(audio=audio_file, **options)
        
```

Also tried this:

```
import os
os.environ["ORT_DISABLE_CPU_AFFINITY"] = "1"
os.environ["OMP_NUM_THREADS"] = "4"
os.environ["OPENBLAS_NUM_THREADS"] = "4"
os.environ["MKL_NUM_THREADS"] = "4"
os.environ["VECLIB_MAXIMUM_THREADS"] = "4"
os.environ["NUMEXPR_NUM_THREADS"] = "4"

```

But to no avail. Any suggestions? Below the crash log.

> Loading large-v2 model...
> Done loading large-v2 model, took: 75.503 seconds
> Starting transcribing
> INFO:faster_whisper:Processing audio with duration 03:25.706
> 2024-11-22 19:33:53.322733977 [E:onnxruntime:Default, env.cc:234 ThreadMain] pthread_setaffinity_np failed for thread: 785, index: 1, mask: {2, }, error code: 22 error msg: Invalid argument. Specify the number of threads explicitly so the affinity is not set.
> INFO:faster_whisper:VAD filter removed 00:19.722 of audio
> DEBUG:faster_whisper:VAD filter kept the following audio segments: [00:00.048 -> 01:07.440], [01:07.984 -> 03:06.576]
> 0%|          | 0/185.98 [00:00<?, ?seconds/s]DEBUG:faster_whisper:Processing segment at 00:00.000
> Traceback (most recent call last):
[...]

PR #1198, opened 2024-12-10
Fixes OOM Errors - too high RAM usage by VAD

### LINKED ISSUE #1193: OOM when using VAD
Hi, does somebody else experience issues with memory consumption when transcribing audio files containing a lot of speech (~ 4 hours long)? I am running the latest version of faster-whisper in a Kubernetes pod on a g4dn AWS instance. The server has 4 cores, 1 GPU, and 16GB RAM, but the pod is limited to 2 cores. The base image is `pytorch/pytorch:2.5.1-cuda12.4-cudnn9-runtime` and as per [this pinned issue](https://github.com/SYSTRAN/faster-whisper/issues/1086) the installed versions should be compatible:
- python 3.11
- torch 2.5.1+cu124
- ctranslate2 4.5.0
- cuda 12.4
- cudnn 9.1.0.7

The process gets killed during the transcription phase when VAD is enabled. I tried the solution [described here](https://github.com/snakers4/silero-vad/issues/356), but it doesn't help. See the logs attached. Anyone has any idea what could be the cause of the OOM?

[libraries.txt](https://github.com/user-attachments/files/18039471/libraries.txt)
[logs on sigkill.txt](https://github.com/user-attachments/files/18039459/logs.on.sigkill.txt)

Purfview:
fixes #1193 and #1169 
VAD implementation consumes humongous memory amount [original Silero doesn't have this problem]

This PR should fix the OOM problem.
Alt solution could be removing `lru_cache`.

### DIFF
--- a/faster_whisper/vad.py
+++ b/faster_whisper/vad.py
@@ -302,7 +302,15 @@ class SileroVADModel:
 
         batched_audio = batched_audio.reshape(-1, num_samples + context_size_samples)
 
-        encoder_output = self.encoder_session.run(None, {"input": batched_audio})[0]
+        batch_process_size = 10000
+        num_segments = batched_audio.shape[0]
+        encoder_outputs = []
+        for start in range(0, num_segments, batch_process_size):
+            end = min(start + batch_process_size, num_segments)
+            encoder_output = self.encoder_session.run(None, {"input": batched_audio[start:end]})[0]
+            encoder_outputs.append(encoder_output)
+
+        encoder_output = np.concatenate(encoder_outputs, axis=0)
         encoder_output = encoder_output.reshape(batch_size, -1, 128)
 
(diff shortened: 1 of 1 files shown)


MahmoudAshraf97:
Thanks, for reference https://github.com/microsoft/onnxruntime/issues/11627

BTW, there is still room for OOM on smaller ram systems caused by this:
https://github.com/SYSTRAN/faster-whisper/blob/8327d8cc647266ed66f6cd878cf97eccface7351/faster_whisper/vad.py#L302-L305
the input here is `(num_segments, 576)` and `num_segments` depends on the audio length and it's unbounded, we should replace this with a for loop over `num_segments` with a large batch size (10k for example or the lowest number where we start to lose speed), this is helpful for very long audios

PR #1239, opened 2025-02-07
Update cudnn semantic version in README to avoid performance issue

kkew3:
I suggest changing the version spec of `nvidia-cublas-cu12` and `nvidia-cudnn-cu12`. Using existing one:

```bash
pip install nvidia-cublas-cu12 nvidia-cudnn-cu12==9.*
```

on 2025-02-07 gives

```
nvidia-cublas-cu12==12.8.3.14
nvidia-cudnn-cu12==9.7.1.26
```

which results in cuda running 30x slower than cpu on my Ubuntu 22.04, with a 6 seconds wav audio.

After I downgrade them to:

```
nvidia-cublas-cu12==12.4.5.8
nvidia-cudnn-cu12==9.1.0.70
```

which is what `torch==2.6.0` currently depends on, everything works fine. Then, I go through a thorough grid search over recent cublas and cudnn versions, and find that all cudnn after 9.6 (inclusive) exhibit the aforementioned performance issue.

I didn't study the lower bound, though. Hence, I make it compatible with current semantic version, i.e. `>=9`).

That's why I make this change.

Thank you so much!

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ The libraries (cuBLAS, cuDNN) are installed in this official NVIDIA CUDA Docker 
 On Linux these libraries can be installed with `pip`. Note that `LD_LIBRARY_PATH` must be set before launching Python.
 
 ```bash
-pip install nvidia-cublas-cu12 nvidia-cudnn-cu12==9.*
+pip install nvidia-cublas-cu12 'nvidia-cudnn-cu12<9.6,>=9'
 
 export LD_LIBRARY_PATH=`python3 -c 'import os; import nvidia.cublas.lib; import nvidia.cudnn.lib; print(os.path.dirname(nvidia.cublas.lib.__file__) + ":" + os.path.dirname(nvidia.cudnn.lib.__file__))'`
 ```


MahmoudAshraf97:
Hello, we just suggest the major version for reference, since this project does not depend on the PYPI cudnn installation we cannot pin the version to anything except the major version

kkew3:
Hello. Thank you so much for your reply!

IMHO it would still be better to make

> we just suggest the major version for reference

explicit in README, so that users will know they may need to adjust the cublas and/or cudnn version accordingly, depending on their own dev environments. It could be somewhat confusing to simply state `pip install nvidia-cublas-cu12 nvidia-cudnn-cu12==9.*` in README, as this command may not work out-of-the-box for every user.

MahmoudAshraf97:
unfortunately, tuning the cudnn version is not an option for most users, especially that it's closely tied with with other packages that might exist in the same environment such as pytorch which is only compatible with a single cudnn version. I know CUDA dependencies are a headache to manage that's why they are out of the scope of the requirements even though they are required, simply because there's no one solution fits all here so we leave it up to the users to decide

kkew3:
I got your point. Thank you so much!

Purfview:
> with a 6 seconds wav audio

What about long audios, an hour and longer?

kkew3:
@Purfview I recently got back to my project using faster-whisper and revisited this issue.

First I need to mention that the issue still persists for `faster-whisper==1.2.1`. I recorded the test code and test audio, which are attached below. Although I didn't delve into the case where the audio lasts for an hour or longer, I believe that case should behave as worse; see the result analysis below.

# Attachment

The test code:

```python
# Usage:
# 1. uv sync
# 2. ./configure ; Write LD_LIBRARY_PATH=xxx to .env file.
# 3. uv run --env-file .env <this-file> <args>...

import time
import argparse

DEFAULT_MODEL = "mobiuslabsgmbh/faster-whisper-large-v3-turbo"

parser = argparse.ArgumentParser(description="Minimal transcription example.")
parser.add_argument("-m", "--model", default=DEFAULT_MODEL)
parser.add_argument("-a", "--audio", default="sample_audio.mp3")
parser.add_argument("--cpu", action="store_true")
args = parser.parse_args()

from faster_whisper import WhisperModel  # noqa: E402


tic = time.perf_counter()
model = WhisperModel(
    args.model,
    device="cpu" if args.cpu else "cuda",
    # On cpu, if using float16, then:
    #   ValueError: Requested float16 compute type, but the target device or
    #   backend do not support efficient float16 computation.
    compute_type="float32" if args.cpu else "float16",
    local_files_only=True,
)
toc = time.perf_counter()
print(f"## Took {toc - tic:.1f} sec")
tic = toc
segments, info = model.transcribe(
    args.audio,
    beam_size=5,
    language="en",
    condition_on_previous_text=False,
)
print(f"Duration: {info.duration}")
for seg in segments:
    print(f"[{seg.start:.2f} -> {seg.end:.2f}] {seg.text}")
toc = time.perf_counter()
print(f"## Took {toc - tic:.1f} sec")
```

The test audio (credit: the first six seconds of the YouTube video <https://youtu.be/lXUZvyajciY?si=fvsmVpiJIkqooIqW>):

[sample_audio.mp3](https://github.com/user-attachments/files/23558285/sample_audio.mp3)

# Result analysis

**With `nvidia-cudnn-cu12<9.6,>=9`**, the test code output:

```
## Took 0.9 sec
Duration: 6.8
[0.00 -> 4.40]  reinforcement learning is terrible it just so happens that everything that we
[4.40 -> 7.40]  had before it is much worse
## Took 7.8 sec
```

Note:

- Took 0.9 seconds to load the whisper model from disk.
- Duration of the audio was detected to last for 6.8 seconds, which is correct.
- The transcripts are correct.
- Took 7.8 seconds to transcribe.

**With `nvidia-cudnn-cu12>=9.6` (e.g. `nvidia-cudnn-cu12==9.6.0.74`)**, the test code output:

```
## Took 1.0 sec
Duration: 6.8
[0.00 -> 6.80] !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
## Took 31.9 sec
```

We see that not only it took much longer time to complete, but also the transcript is garbage.

**Conclusion**:

I can't identify whether this is an upstream issue or not, but clearly something went wrong beyond taking too much time to complete.

kkew3:
Thank you so much! Following your suggestions there it works perfectly.

Summary table for future reference:

GPU: T1000

| cudnn version bounds | compute dtype | time taken to transcribe | transcript |
|---------------|---------------|-------------------------|------------|
| `>=9,<9.6`| `float16` | 7.9 seconds | correct |
| `>=9.6,<10` | `float16` | 31.9 seconds | garbage |
| `>=9.6,<10` | `float32` | 2.3 seconds | correct |

Issue #364, opened 2023-07-20
VAD is relatively slow


AlexandderGorodetski:
Hello guys,

I am using VAD of faster whisper using following commands. I found that on TedLium benchmark transcribing VAD takes 8% of time and 92% takes transcribing. I would prefer to decrease time of VAD so that it will not take more than 1%. Is it somehow possible to optimize VAD procedure in terms of real time?? Maybe it is possible to run VAD on several CPU's? BTW, I see that VAD is running on CPU, is it possible to run it somehow on GPU?

```python
# VAD
audio_buffer = decode_audio(audio_filename,
                                sampling_rate=whisper_sampling_rate)

# Get the speech chunks in the given audio buffer, and create a reduced audio buffer that contains only speech.    
speech_chunks = get_speech_timestamps(audio_buffer)
vad_audio_buffer = collect_chunks(audio_buffer, speech_chunks)

# Transribe the reduced audio buffer.
init_segments, _ = whisper_model.transcribe(vad_audio_buffer, language=language_code, beam_size=beam_size)

# Restore the true time-stamps for the segments.
segments = restore_speech_timestamps(init_segments, speech_chunks, whisper_sampling_rate)
```

archive-r:
Lowering the window_size_sample value may help.

https://github.com/snakers4/silero-vad/issues/322#issuecomment-1519015503

guillaumekln:
The VAD model is also run on a single CPU core:

https://github.com/guillaumekln/faster-whisper/blob/e786e26f75f49b7d638412f3bf2b2b75a9c3c9e8/faster_whisper/vad.py#L254-L255

Can you try changing these values and see how they impact the performance?

phineas-pta:
u can make vad run on gpu

1. install dependencies

```
pip uninstall onnxruntime
pip install onnxruntime-gpu
```

2. edit code

in `vad.py` line 260 change `CPUExecutionProvider` to `CUDAExecutionProvider`

phineas-pta:
u can make vad run on gpu

1. install dependencies

```
pip uninstall onnxruntime
pip install onnxruntime-gpu
```

2. edit code

in `vad.py` line 253-262 replace with

```python
        opts = onnxruntime.SessionOptions()
        opts.log_severity_level = 4
        opts.graph_optimization_level = onnxruntime.GraphOptimizationLevel.ORT_ENABLE_BASIC
[...]

PR #1262, opened 2025-03-12
Decouple onnxruntime cpu and gpu requirements

prassanna-ravishankar:
This has been discussed before in https://github.com/microsoft/onnxruntime/issues/7313 and https://github.com/SYSTRAN/faster-whisper/pull/499

#### Problem description

- Let's package `xyz` (meant to be GPU accelerated) depends on faster-whisper
- `xyz` explicitly lists onnxruntime-gpu in their dependencies
- `xyz` uses a package manager with a lockfile setup (like `poetry` or `uv`)
- Installing package `xyz` ends up installing both `onnxruntime` and `onnxruntime-gpu` in its environment

This is undesirable (https://github.com/SYSTRAN/faster-whisper/issues/364#issuecomment-1645272083) , and requires `onnxruntime` to be manually removed, and `onnxruntime-gpu` to manually be reinstalled. 

#### Proposed solution

- Break out the CPU and GPU requirements as explicit extras. This is a **breaking change**
- `pip install faster-whisper` will **not** install onnxruntime in anyflavour.
- Instead, one will have to `pip install faster-whisper[cpu]` or `pip install faster-whisper[gpu]` 

Happy to modify and iterate on this PR as much as required, but due to onnxruntime's weird behaviour (onnxruntime-gpu incompatible with onnxruntime), this is causing many downstream projects to "invent" their own way of dealing with non-standard installs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Install module
         run: |
           pip install wheel
-          pip install -e .[dev]
+          pip install -e .[dev,cpu]
 
       - name: Check code format with Black
         run: |
@@ -55,7 +55,7 @@ jobs:
       - name: Install module
         run: |
           pip install wheel
-          pip install -e .[dev]
+          pip install -e .[dev,cpu]
 
       - name: Run pytest
         run: |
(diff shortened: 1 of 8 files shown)


MahmoudAshraf97:
@Purfview I'd appreciate your input on this

Purfview:
> @Purfview I'd appreciate your input on this

I think this breaks more than it fixes.

PR #1274, opened 2025-03-28
Add language output to each segment

erichaydel:
This adds the already existing tokenizer language to the outputs of each segment. 

This allows faster-whisper to be used as a language detect in multilingual mode without any extra processing burden in both regular and batched mode, and uses the primary language if multilingual mode is off.

```python
segments, _ = model.transcribe("audio.mp3", multilingual=True)

for segment in segments:
    print("[%.2fs -> %.2fs] (%s) %s" % (segment.start, segment.end, segment.language, segment.text))
```
>[0.02s -> 21.42s] (es)  Esta mañana, la primera comisión continuará adoptando medidas sobre los proyectos de resolución y de decisión restantes.


RyuseiMigita:
I expecting this function release. Any idea about if this would be released in near the future?

PR #1291, opened 2025-04-24
Add support AsyncBatchedInferencePipeline

sh1man999:


Purfview:
Please keep the comments in English (there are multiple comments in Russian).

sh1man999:
> This introduces a lot of code duplication, please inherit the base class of `BatchedInferencePipeline` and modify the needed functions only, also do not use `asyncio.to_thread`

what other test is needed. I have already added the test_async_batched_transcribe test

Scott-Hickmann-Sanas:
Why was this not merged? This would be a very helpful addition to this library.

MahmoudAshraf97:
It's indeed a very helpful addition, but this is a very low quality PR for such an important feature, I'm willing to accept contributions if someone is willing to put enough effort in it and not just slap a bunch of AI authored commits

PR #1292, opened 2025-04-26
feat: allow passing specific revision to download

felixmosh:
This feature allows to specify which revision of the model do you want to download & use

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -596,6 +596,7 @@ class WhisperModel:
         download_root: Optional[str] = None,
         local_files_only: bool = False,
         files: dict = None,
+        revision: Optional[str] = None,
         **model_kwargs,
     ):
         """Initializes the Whisper model.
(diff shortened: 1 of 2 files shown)


MahmoudAshraf97:
Hello and thanks for your contribution, I have a question, when is this useful? because most whisper models uploaded have only a single revision that includes the model, other revisions are mostly updates to misc files, and also when is the use of an older revision needed?

felixmosh:
Hi, there are models that using revisions of improvements of the fine-tuning Whisper model. 

I want to keep in the code (& git) the version we are currently using.

MahmoudAshraf97:
understood, can you make sure the formatting adheres to the repo standard and the check are passing? check the contributors guide for steps

felixmosh:
Done

MahmoudAshraf97 (approved):

MahmoudAshraf97:
You can install from git directly, I don't think this is worth it's own release as it's neither a bug fix nor a feature

PR #1297, opened 2025-05-10
Remove Silence in Batched transcription

MahmoudAshraf97:
The current VAD implementation in batched transcription is only used for segmentation, silence is only removed at segments boundaries, for example, if we have a speech segment from 1 to 3 and another from 9 to 10, the resulting segment will be from 1 to 10, including a large silence period from 3 to 9 which is prone to hallucinations

This PR concatenates speech only until the desirable chunk size is reached and ignoring the silence in-between thus reducing hallucinations in batched mode
Since the speech is condensed, more words per segment should be expected in the transcript, potential slight speedup is also expected due to less segments for long files

Results after increasing `min_silence_duration_ms` to 800:
| Model               | Before WER | After WER  |
|---------------------|------------|------------|
| tiny.en             | **15.063** | 15.088     |
| tiny                | **21.390** | 22.112     |
| base.en             | 13.816 | **13.597**     |
| base                | 17.251 | **16.735**     |
| small.en            | **12.617**     | 12.649 |
| small               | 16.088 | **16.050**    |
| medium.en           | 12.894     | **12.786** |
| medium              | 15.593 | **15.218**     |
| large-v1            | 19.590     | **16.121** |
| large-v2            | 15.148     | **14.103** |
| large-v3            | 15.997     | **15.549** |
| large-v3-turbo      | 14.044     | **13.561** |
| distil-small.en     | 13.918     | **13.784** |
| distil-medium.en    | 13.972 | **13.916**     |
| distil-large-v2     | **13.574** | 13.595     |
| distil-large-v3     | 13.533     | **13.512** |

These figures can be reproduced by running `benchmarks/yt_commons.py`

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -25,7 +25,6 @@ from faster_whisper.vad import (
     VadOptions,
     collect_chunks,
     get_speech_timestamps,
-    merge_segments,
 )
 
 
@@ -125,7 +124,7 @@ class BatchedInferencePipeline:
         segmented_outputs = []
         segment_sizes = []
         for chunk_metadata, output in zip(chunks_metadata, outputs):
-            duration = chunk_metadata["end_time"] - chunk_metadata["start_time"]
+            duration = chunk_metadata["duration"]
             segment_size = int(ceil(duration) * self.model.frames_per_second)
             segment_sizes.append(segment_size)
             (
@@ -135,7 +134,7 @@ class BatchedInferencePipeline:
             ) = self.model._split_segments_by_timestamps(
                 tokenizer=tokenizer,
                 tokens=output["tokens"],
-                time_offset=chunk_metadata["start_time"],
+                time_offset=chunk_metadata["offset"],
                 segment_size=segment_size,
                 segment_duration=duration,
                 seek=0,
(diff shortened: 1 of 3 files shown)


MahmoudAshraf97:
@Purfview I'd appreciate your review and testing

Jiltseb:
@MahmoudAshraf97 Indeed, it can reduce the hallucinations. However, if each voiced region returned by VAD is treated as a single segment and padded to 3sec for whisper, we are indeed increasing the number of such segments and effectively not packing in multiple segments in a single GPU. This can ruin efficiency. 

We can adjust the merging logic to ensure chunks are filled with as many voiced regions as possible without exceeding the 30-second limit. We will have to maintain a mapping of original segment indices and their timings for later retrieval. This way, we make sure that multiple segments are kept in a single input (30 sec) to the whisper encoder and at the same time avoid current hallucinations.
@MahmoudAshraf97 Thoughts?

MahmoudAshraf97:
Hello @Jiltseb, this is already the case, we pack regions until the total duration reaches 30s, and after inference, we map the predicted timings to the original ones

Jiltseb (approved):
👍 LGTM!

PR #1299, opened 2025-05-15
Do not encode supplied `initial_prompt` if it is not a string

jzyeezy:
The `BatchedInferencePipeline#transcribe` method [technically supports](https://github.com/SYSTRAN/faster-whisper/blob/700584b2e6c15825c73e232f323e4b910257fa1c/faster_whisper/transcribe.py#L280) passing an already encoded tokenized list as the `initial_prompt`. The code path should only try to encode the provided prompt if it's a string.

This fixes the bug where faster_whisper will fail due to: `TypeError: TextInputSequence must be str` when passing a List of integers as the `initial_prompt`.


MahmoudAshraf97:
Also, even if the typing hints support, Why would you use a tokenized string instead of a string? note that different whisper models use different tokenizers so the tokenized string is not generalizable
I prefer if we change the typing hint to just accept `str` instead

PR #1305, opened 2025-05-23
Make prefix work for batched inference

Scott-Hickmann-Sanas:
Prefix is ignored during batched inference, resulting in a different behavior than when running faster whisper without batched inference. This PR aims to be a simple fix for that.


MahmoudAshraf97:
Prefix means a part of the actual transcription that whisper should complete, this is not relevant for batched transcription unless all segments in an audio file begin with the same prefix, which is an almost non-existent case

Issue #1319, opened 2025-07-08
BatchedInferencePipeline - nocaptions


mmichelli:
When I use the BatchedInferencePipeline I get <|nocaptions|>. 

The person is speaking Norwegian -- without a dialect, and the language is set to Nynorsk. 
 - Nynorsk is primarily a written form of Norwegian, not a spoken language, but it's usually Norwegian with a dialect.

## BatchedInferencePipeline
```python
model = WhisperModel("NbAiLab/nb-whisper-large", device="cuda" )
batched_model = BatchedInferencePipeline(model=model)

segments, info = batched_model.transcribe("../data/audio_1_extract_min4to12.mp3", vad_filter=True ,  language="nn", word_timestamps=True,  task="transcribe" ) 
 
def seconds_to_hms(seconds):
    hours, remainder = divmod(seconds, 3600)
    minutes, seconds = divmod(remainder, 60)
    return f"{int(hours):02d}:{int(minutes):02d}:{int(seconds):02d}"
 
for segment in segments:
    print(f"{seconds_to_hms(segment.start)} -> {seconds_to_hms(segment.end)} {segment.text}")

```
00:03:33 -> 00:03:37 <|nocaptions|>
00:04:03 -> 00:04:06 <|nocaptions|>
00:04:31 -> 00:04:34 <|nocaptions|>

## WhisperModel
```python
model = WhisperModel("NbAiLab/nb-whisper-large", device="cuda" )

segments, info = model.transcribe("../data/audio_1_extract_min4to12.mp3", vad_filter=True ,  language="nn", word_timestamps=True,  task="transcribe" ) 
 
for segment in segments:
    print(f"{seconds_to_hms(segment.start)} -> {seconds_to_hms(segment.end)} {segment.text}")

```

00:03:30 -> 00:03:34  Ansatte legar ved legetenesta i Vinje. Det er oversikta.
00:03:34 -> 00:03:36  I dag har vi åtte fastlegar.
00:03:37 -> 00:03:42  To delar på lista i Edeland er fysisk til stades torsdagar og fredagar.
00:03:43 -> 00:03:48  Begge har 100 % stilling, men dei er der ute to dagar i veka.

MahmoudAshraf97:
Can you test with this PR #1297 ?

mmichelli:
Looks like the problem is missing suppress tokens  'no', 'ca', 'ptions'

```py
[...]

PR #1338, opened 2025-08-12
Fix: Prevent <|nocaptions|> tokens in BatchedInferencePipeline

mmichelli:
https://github.com/SYSTRAN/faster-whisper/issues/1319#issuecomment-3153705309

- Add nocaptions component tokens [1771, 496, 9799] to suppress_tokens list
- Add segment filtering to remove any remaining <|nocaptions|> segments
- Resolves issue where BatchedInferencePipeline would generate malformed special tokens during periods of silence or low-confidence transcription
- Includes comprehensive tests to verify the fix

The issue occurred because while bracket tokens ('<', '|', '>') were already suppressed, the content tokens ('no', 'ca', 'ptions') were not, leading to partial token generation that formed complete <|nocaptions|> tags in the output.

Files changed:
- faster_whisper/transcribe.py: Core fix implementation
- test_nocaptions_comprehensive.py: Comprehensive test suite
- tests/test_nocaptions_fix.py: Unit tests

### DIFF
--- a/PR_DESCRIPTION.md
+++ b/PR_DESCRIPTION.md
@@ -1,0 +1,73 @@
+# Fix: Prevent `<|nocaptions|>` tokens in BatchedInferencePipeline
+
+## Problem
+
+When using `BatchedInferencePipeline` for transcription, malformed `<|nocaptions|>` special tokens were appearing in the output during periods of silence or low-confidence transcription. This issue did **not** occur with the regular `WhisperModel.transcribe()` method.
+
+### Example of the issue:
+```
+00:05:36 -> 00:06:06  <|nocaptions|>
+00:06:06 -> 00:06:37  <|nocaptions|>
+00:06:37 -> 00:07:09  <|nocaptions|>
+```
+
+## Root Cause
+
+The issue was caused by **partial token suppression** in the `get_suppressed_tokens()` function:
+
+1. **Bracket tokens suppressed**: The tokens `[27, 91, 29]` representing `['<', '|', '>']` were already in the `non_speech_tokens` list and properly suppressed
+2. **Content tokens NOT suppressed**: The tokens `[1771, 496, 9799]` representing `['no', 'ca', 'ptions']` were **not** being suppressed
+3. **Result**: The model could still generate "nocaptions" content, and when combined with any remaining bracket tokens, formed complete `<|nocaptions|>` tags
+
+## Solution
+
+This PR implements a two-layer fix:
+
+### 1. Token-level suppression (Primary fix)
+```python
+# Add nocaptions component tokens to prevent malformed <|nocaptions|> generation
+suppress_tokens.extend([1771, 496, 9799])  # 'no', 'ca', 'ptions'
+```
+
+### 2. Segment-level filtering (Safety net)
+```python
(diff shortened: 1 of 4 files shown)


PR_DESCRIPTION.md
MahmoudAshraf97, line 1: Remove this

test_nocaptions_comprehensive.py
MahmoudAshraf97, line 1: remove this

tests/test_nocaptions_fix.py
MahmoudAshraf97, line 1: remove this

MahmoudAshraf97 (changes requested):
Hello, `<|nocaptions|>` is a single token not a multiple token segment
what needs to be done here is add a `no_speech` attribute to the tokenizer which should be equal to `model.hf_tokenizer.token_to_id("<|nocaptions|>")` and add this to the suppressed tokens

mmichelli:
I've updated the PR to implement the single token approach as suggested by @MahmoudAshraf97.

**Changes made:**
- Added `no_speech` property to Tokenizer class using `tokenizer.token_to_id("<|nocaptions|>")`
- Added null checks to handle models without this token
- Updated `get_suppressed_tokens` to conditionally include the `no_speech` token
- Updated tests with correct token ID expectations

**Testing:**
- All tokeniser tests pass
- Works with different model variants (tiny, large-v3, etc.)
- No breaking changes

mmichelli:
I thought it was possible that the `<|nocaptions|>` were missing.

For example:
https://huggingface.co/NbAiLab/nb-whisper-large/blob/main/tokenizer.json

```python
from faster_whisper import WhisperModel

model = WhisperModel("NbAiLab/nb-whisper-large", device="cuda")
nocaptions_token_id = model.hf_tokenizer.token_to_id("<|nocaptions|>")
print(f"<|nocaptions|> token ID: {nocaptions_token_id}")
```

**<|nocaptions|> token ID: None**

RafalCer:
Hey, @mmichelli I ran into the same issue with NB whisper, i.e. constant `<|nocaptions|>` on segments with background noise or multiple people talking at the same time. 

As you noted in your previous comment, the token (`<|nocaptions|>`) is not available in the tokenizer of the model. You can print out the token indices to see that whenever NB Whisper's output is `<|nocaptions|>` , the model actually outputs a sequence of tokens to construct it, as you can see in the example below:

```
Subsegment tokens: [27, 91, 1771, 496, 9799, 91, 29]
Subsegment text: <|nocaptions|>
```

I tried adding `<|nocaptions|>` to vocabulary, but ultimately solved the issue for me was adding  `<` id, which is 27, to `suppress_ids_begin` in `config.json`. To the best of my understanding, this way all outputs starting with `<` get suppressed and NB whisper outputs its second best guess for the sequence. From my quick evaluation, the second best guess for what would otherwise be `<|nocaptions|>` is almost always correct. Plus, this way I have way less gaps in the transcript.

MahmoudAshraf97:
This is a problem with the finetune itself, it's not a case we should handle in this PR, and users of this model will be able to pass any extra tokens they want to suppress to `transcribe` function

PR #1341, opened 2025-08-14
Adds a quick and dirty Gradio frontend

gravis778:
Silently exits after creating transcript, but it DOES create the transcript in the app directory


MahmoudAshraf97:
Hello and thanks for the contribution, this is out of scope of this repo, please implement it as a separate project

Issue #1348, opened 2025-08-20
Multilingual_transcription test failing with the German transcription.


mmichelli:
The test_multilingual_transcription test in tests/test_transcribe.py is currently failing with an AssertionError. This is caused by a typo in the expected German transcription.

 ### To Reproduce:
  1. Install the test dependencies.
  2. Run pytest.

 ### Error Details:
```bash
   1 E       AssertionError: assert ' Jedem, der ...u übertragen.' == ' Jedem, der ...u übertragen.'
   2 E         Skipping 414 identical leading characters in diff, use -v to show
   3 E         - nterfolgen den Bedingungen anderen zu übertragen.
   4 E         ?           -
   5 E         + nterfolgenden Bedingungen anderen zu übertragen.
   6
   7 tests/test_transcribe.py:186: AssertionError

```

 ### Suggested Fix:
 The expected German text in tests/test_transcribe.py appears to have a typo. The word unterfolgen should likely be unterfolgenden.

PR #1349, opened 2025-08-20
Fix typo in multilingual test

mmichelli:
Fixes a typo in the German transcription of the multilingual test.

https://github.com/SYSTRAN/faster-whisper/issues/1348

I don't know German so I asked an LLM: 

The second option, **"unter folgenden Bedingungen anderen zu übertragen"**, is the most correct.

Here's a breakdown of why:

* **"Unter"** is a two-way preposition in German, meaning it can take either the dative or accusative case. In this context, where it means "under" or "subject to" certain conditions, it requires the **dative case**.
* **"Bedingungen"** is a plural noun.
* The adjective **"folgend"** (following) comes before the noun and must therefore be declined according to the dative plural case.

The correct dative plural declension for an adjective like "folgend" is to add an **-en** ending. Therefore, "folgenden" is the grammatically correct form.

The first option, "unter folgen den Bedingungen," incorrectly separates the adjective ending and uses an incorrect declension.


mmichelli:
Ok, I see this breaks the bests in Google. 

Is the difference with or without a GPU?

PR #1357, opened 2025-09-05
Upgrade to Silero-Vad V6

Pastells:
481712359-c485da29-5814-409d-9a94-e5b0f582da17

https://github.com/snakers4/silero-vad/releases/tag/v6.0

- Replace ONNX file with V6 version
- Instead of an encoder and a decoder it is a single file


Pastells:
I'm not sure how you got the v5 encoder and decoder. Did you manually split them?
Anyway, #1365 also uses a single model, but has done some other changes, so it's probably better to merge that than my commit.

Issue #1330, opened 2025-07-31
how to transcrib multiple short audio (10-30s) parallel. use max Gpu


hoangquocvietbro:
I have an audio file and a timestamp for each sentence. I’ve already split the audio into multiple clips with durations ranging from 10 to 30 seconds. However, I don’t know how to transcribe them in parallel, similar to using batch_size with a long audio file. I need to transcribe them according to the existing timestamps and fully utilize the GPU for maximum speed. What should I do?

I tried using multiprocessing, but it seems that model.transcribe() still runs sequentially and uses very little GPU.

Issue #1264, opened 2025-03-15
Batch with multi audio input.


MM-0712:
Hi,
I see the batched inference is for the long audio. If the input is multi-short audio, could it support the batched inference?
Thank you in advance.

MahmoudAshraf97:
Hi, It's doable but I don't have the capacity to implement it, PRs are welcome though

shushenghong:
maybe https://github.com/SYSTRAN/faster-whisper/issues/1177?

Issue #1192, opened 2024-12-06
 Lack of GPU Parallelism for Real-Time Server Using Faster-Whisper


dariopellegrino00:
Hi, i'm currently working on my thesis, which involves building a real-time transcription server using the `whisper-streaming` project and `faster-whisper` for the ASR backend. The server is deployed on an RTX 6000 Ada GPU, but I am struggling to achieve proper GPU parallelism.
I am relatively new to using Whisper and have only recently started using Python. I appreciate your patience and any guidance you can provide!

#### I Tried

**Multiple Models on Multiple Threads**:
   - I instantiated multiple `WhisperModel` instances (one per thread) and assigned each client to its own model. While this approach works for a few clients, performance degrades significantly beyond ~8 clients, regardless of the model size. Visually what seems to be happening to me is that the models are competing with each other for the entire GPU resources.

**Single Shared Model with `num_workers`**:
   - I shared a single `WhisperModel` instance among multiple threads and used the `num_workers` parameter to enable concurrent processing. This approach also works well initially but similarly fails to handle more than ~8 clients effectively, again with the same issues. 
   
1. Is there a way to achieve true GPU parallelism for multiple audio sources on a single GPU using `faster-whisper`?
2. Does the `num_workers` parameter have any impact on GPU-based inference, or is it exclusively for CPU execution?
3. Are there recommended configurations or best practices for maximizing GPU utilization in scenarios with multiple concurrent audio streams?

Any advice or clarification would be greatly appreciated. Thank you for your amazing work on this project!

heimoshuiyu:
Hello, I wrote a very simple FastAPI script to run the faster whisper module. The script is available here. <https://github.com/heimoshuiyu/whisper-fastapi>

I use Docker to deploy my service. I have 4 RTX 4070 Ti Super GPUs, and I deploy 2 services on each GPU. So in total, I have 8 services. Each service is mapped to one client, and I set up the Grafana GPU monitor, which indicates all GPUs are utilized at 100%.

I am using the large v2 module, and almost all my transcription tasks are longer than 1 to 3 minutes of audio. The feature extraction preprocessing part isn't wasting too much GPU time, I think. In my case, two services per GPU is enough. You might consider using more service per GPU if you are transcribing shorter audio.

dariopellegrino00:
Thank you very much for the response. I will take a look at your solution as soon as I can.

MahmoudAshraf97:
[...]

Issue #1177, opened 2024-11-27
Batched inference on multiple audios


Deep-unlearning:
Hello,

We are currently looking to add faster-whisper into the [Open ASR Leaderboard](https://huggingface.co/spaces/hf-audio/open_asr_leaderboard). 

Here is the script we are using to run the evals: https://github.com/huggingface/open_asr_leaderboard/tree/main/ctranslate2

We noticed that I does not natively support batched inference on multiple audios which makes the RTFx significantly lower that the orignal whisper which are evaluated with a batch_size 64.

Is there something that we are doing wrong ? 

Thanks

Vaibhavs10:
cc: @MahmoudAshraf97 maybe - would be great to get another set of eyes on the eval scripts.

MahmoudAshraf97:
unfortunately it doesn't yet, the batching we support is on a single file that is segmented using VAD and AFAIK, OpenASR leaderboard discourages the usage of VAD.
If you guys are willing to monkey-patch, CT2 inference can be inserted into any library that already supports multiple files such as transformers or the original one, so that the heavy lifting is done by CT2 and the pre/post processing logic is done by any other library

Vaibhavs10:
Hi @MahmoudAshraf97 - thanks for the response. do you have a sample code for how this would look like with transformers? or can you please point us to the right documentation? 🙏

MahmoudAshraf97:
You'll have to replace [This](https://github.com/huggingface/transformers/blob/c7a109ec81fb38fdd77d1d0543a9233bc6bf4d09/src/transformers/models/whisper/generation_whisper.py#L847-L857) with https://github.com/SYSTRAN/faster-whisper/blob/8327d8cc647266ed66f6cd878cf97eccface7351/faster_whisper/transcribe.py#L223-L237

Just some quick tips: 
1. CT2 has a limitation where `<|startoftranscript|>` token must be in the same position for the whole batch, and padding is not officially supported yet (there is a PR for that if you are willing to use a custom build), so you will have to turn off `condition_on_previous_text`
2. CT2 is extremely slow when batched and `<|notimestamps|>` token is not present, which I think a requirement for accurate longform generation, I don't know what is the solution for that but keep it in mind
3. If you want to test batching without VAD, I suggest you use the HF longform chunking implementation which can be engaged when `chunk_length_s` is not None in pipeline class

PR #1359, opened 2025-09-08
Work in progress for batching with multiple audio files [WIP]

j-silv:
Concerns issues #915, #1177, #1264, #1192, #840, #1330, #59, #66

This work continues where PR #1302 left off. The goal is to transcribe multiple audio files truly in parallel and increase GPU throughput.

**This PR is not done yet**, but I wanted to give a preview and discuss some of the choices I have made so far.

## Here's an overview of what I did so far

1. Batch a list of files (`batch_audio_files()`) and pad each decoded audio to max length within that batch. Users can also just pass in an already batched numpy array, where it's assumed that the first dimension corresponds to different audio file content.

2. I modified the `transcribe` function in `BatchedInferencePipeline` to support batching across multiple files, while still enabling batching across an individual large file. This was a little tricky, but essentially we keep track of how many chunks correspond to each audio file within a batch (added data member `num_chunks` to `TranscriptionInfo`). Then we flatten everything by stacking the features on top of each other. Finally, we pass it into `_batched_segments_generator`, which doesn't have to change since it already supports batching.

3. In addition to performing parallel transcription, I also set up batched VAD. The post-inference algorithm is ran for each individual audio file within the batch and the results are appended together.

## Some of the decisions I took:

1. There are some workarounds in the issues I linked to support multiple audio files. For example, chunking the decoded audio, padding it, and then manually passing in some clip or VAD parameters. I think the main downside to this is that you'd have to do a lot more work to be able to run VAD on the individual audio clips in batch. Plus there's much more work for the user. 

2. Right now, the user is responsible for regrouping the flattened segments array, such that the transcriptions for each audio file are independent. I return a list of generators (one for each batch) in order to do this. I then use `num_chunks` to know when to stop processing segments for the particular audio file. See `test_transcribe::test_batched_transcribe_many` for more info on this. I wasn't sure the best way to approach this, since the original code yields a single segment at a time. Happy to discuss if there are better methods. Ideally I would think the user should not have to worry about regrouping a flat segment array...

3. It is assumed that all audio files are in the same language. In theory you could support batching different languages by instantiating a new `Tokenizer` within the batch loop with a language indexed from an array (provided by the user or detected automatically). I'm just not sure how common that use-case is.

## Some pending todos:

1. I only handled batch transcription when the language is specified, VAD is enabled, and `clip_timestamps` is **not** provided. I have to make some modifications such that other conditional branches in `transcribe` work (multiple tests are failing because of this).

2. I have not exhaustively tested the batching to make sure the results are as expected. I think an additional test to run is to perform batch transcribe with 2 batched audio files, and then compare the results with calling normal transcribe twice with each of those audio files separately.

3. Run performance tests on GPU to make sure batching results in a speed-up (also make sure utilization goes up as expected)

4. Documentation needs to be rewritten up on how to use this new batch mode

_Please share any concerns or suggestions you have!_

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -106,6 +106,7 @@ class TranscriptionInfo:
     all_language_probs: Optional[List[Tuple[str, float]]]
     transcription_options: TranscriptionOptions
     vad_options: VadOptions
+    num_chunks: Optional[List[int]] = None
 
 
 class BatchedInferencePipeline:
@@ -251,6 +252,47 @@ class BatchedInferencePipeline:
 
         return encoder_output, output
 
+    def batch_audio_files(
+        self,
+        audio_files: Union[str, list[str]],
+        sampling_rate: int = 16000,
+        batch_size: int = 8,
+    ) -> Iterable[Tuple[np.ndarray, list[int]]]:
+        """Batch multiple audio files together similar to PyTorch's collate_fn.
+
+        Arguments:
+            audio_files: Path to a single audio file (or a file-like object), or a list of files.
+            sampling_rate: Resample the audio to this sample rate.
+            batch_size: Max size of a single batch.
+
+        Returns:
+            A generator of numpy arrays with size (batch_size, X) where
+            X is the max audio size in the batch.
+
+            A generator of non-padded valid lengths of each audio file in the batch
+        """
+
+        if not isinstance(audio_files, list):
+            audio_files = [audio_files]
+
+        for i in range(0, len(audio_files), batch_size):
+            batch_audio_files = audio_files[i : i + batch_size]
+
+            audios = []
+            lens = []
+            max_len_in_batch = 0
+            for audio_file in batch_audio_files:
+                audio = decode_audio(audio_file, sampling_rate=sampling_rate)
+                max_len_in_batch = max(max_len_in_batch, len(audio))
+                lens.append(len(audio))
+                audios.append(audio)
+
+            batched_audios = np.stack(
+                [pad_or_trim(audio, length=max_len_in_batch) for audio in audios]
+            )
+
+            yield (batched_audios, lens)
+
     def transcribe(
         self,
         audio: Union[str, BinaryIO, np.ndarray],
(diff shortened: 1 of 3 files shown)


MahmoudAshraf97:
I'll explain the current approaches for batching in both short-form <30s and long-form >30s:

1. Multiple short-form files can be batched without VAD using batched pipeline, no need to work on that case in this PR and leave it for later
2. Multiple long files
a.  Using batched inference, there is no gain here so I wouldn't be concerned with this case wither
b. Using sequential inference, this is where the most speedups are, you should use this as a reference https://github.com/huggingface/transformers/blob/7d57b31e16d5a08fda29578c31fc1924fefc50b2/src/transformers/models/whisper/generation_whisper.py#L796 and should be the sole focus of this PR

BBC-Esq:
> > Hi and thanks for the effort, I guess adding multiple audio batching to `BatchedInferencePipeline` will have a minimal benefit since it already has a very good utilization, I suggest to only focus on the regular transcription since it under utilizes the gpu, there is an implementation for that in the transformers library that you can transfer here with minimal modifications to save efforts
> 
> Would you mind linking to the implementation you had in mind? I get a little loss navigating the transformers library.
> 
> Also if I make changes to the regular transcription, then I would only be supporting batching for cases 3 and 4 below correct?
> 
>     1. x1 where x1 is an audio clip < 30 seconds
> 
>     2. x1 where x1 is an audio clip > 30 seconds
> 
>     3. [x1, x2, x3] where x1, x2, and x3 are audio clips < 30 seconds
> 
>     4. [x1, x2, x3] where x1 is an audio clip > 30 seconds, and x2 and x3 are audio clips < 30 seconds
> 
> 
> In case 2, for example, I believe that the normal transcription calls whisper in 30 second chunks sequentially. So I guess in case 4 we would have a sliding window of 30 seconds across a batch of size N where N corresponds to the number of audio files?
> 
> EDIT: Also I think the main benefit this PR addresses in the `BatchedInferencePipeline` is for case 3. If the user wants to transcribe a bunch of audio files < 30 seconds then they could get processed in parallel using the modifications I proposed. The batched algorithm that exists now only batches audio files > 30 seconds, as I understand it.

Hey @j-silv The WhisperS2T library does what you're describing - i.e. batches multiple files together.  It's my understanding that faster-whisper batches per file and the main difference is that WhisperS2T batches across multiple files.  Each has their pluses and minuses...faster-whisper has some features that WhisperS2T doesn't.  WhisperX is better for timestamps/captions.  But as far as sheer speed when processing numerous files, you can't beat WhisperS2T, especially if there are a lot of shorter files since with faster-whisper the more files there are the more overhead you encounter.  Here's my repo where I use WhisperS2T specifically for multiple files.

https://github.com/BBC-Esq/WhisperS2T-transcriber

PR #1365, opened 2025-09-18
Upgrade to Silero-VAD V6

sssshhhhhh:
Upgrade vad to v6 with an reimplemented graph for speed. 50% faster than the current batched implementation by:

* Slightly faster preprocessing
* Not calculating a useless column in the STFT
* Decoding the entire batch with the LSTM op

V5 model can be found in commit history and matched the current implementation to rtol=1e-5, atol=1e-6.
From testing `intra_op_num_threads = 2` gives a futher 25% boost but anything more is only a couple % faster. Could also run the encoder async on multiple threads which is like 8x faster with enough cores but that's probably overcomplicating it.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,3 @@
-include faster_whisper/assets/silero_encoder_v5.onnx
-include faster_whisper/assets/silero_decoder_v5.onnx
+include faster_whisper/assets/silero_vad_v6.onnx
 include requirements.txt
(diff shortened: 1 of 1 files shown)


MahmoudAshraf97:
Hi and thanks for the contribution, I'm skeptical that you can batch the lstm decoding because each sample needs the previous state to be decoded, the model will work, but the results will be different, can you share the python script you used to build the onnx model?

PR #1379, opened 2025-10-28
VAD: Don't pad if padding is not needed

Purfview:
Avoids appending extra window when audio length is already multiple of window_size_samples


MahmoudAshraf97 (approved):

Purfview:
> Any Idea why the CI fails? it seems unrelated, I also retried it multiple times

Lets check.

MahmoudAshraf97:
I say we leave it as it was, one more window never hurt anyone

PR #1382, opened 2025-10-30
Offload retry logic to hf hub

MahmoudAshraf97:
this simplifies the retry logic introduced in #215 and #228 , newer versions of `huggingface_hub` already handle this logic internally
The motivation for this is that `huggingface_hub` switched from `requests` to `httpx` as their backend, so we either cap the version range or simplify our code to remove the need for `requests`

cc @Purfview

### DIFF
--- a/faster_whisper/utils.py
+++ b/faster_whisper/utils.py
@@ -5,7 +5,6 @@ import re
 from typing import List, Optional, Union
 
 import huggingface_hub
-import requests
 
 from tqdm.auto import tqdm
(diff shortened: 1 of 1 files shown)


Purfview:
I think, adding "requests" is least intrusive.
Your approach reduced code complexity, but restricted huggingface_hub>=0.21

And I wonder why I have huggingface_hub capped at "0.19.4", some lib definitely didn't work with higher version.

PR #1390, opened 2025-11-07
Update Silero-VAD weights to v6.2

Purfview:
- Overall slight quality improvement (no metrics update); 
- Higher stability on OOD / rare / strange / unique data; 
- Significant quality improvements on various known edge cases:
     - Unusual voices
     - Child voices
     - Cartoon voices
     - Muted voices
     - Muted speech
     - Lower quality phone calls

https://github.com/snakers4/silero-vad/releases/tag/v6.2

This is the code I used to reconstruct the onnx model:
https://gist.github.com/MahmoudAshraf97/29f73de73beb8e4549dedb8b5eac9702

Compared VAD's output to the reference - same.

### DIFF
--- a/tests/test_transcribe.py
+++ b/tests/test_transcribe.py
@@ -245,7 +245,7 @@ def test_transcribe_signature():
 
 
 def test_monotonic_timestamps(physcisworks_path):
-    model = WhisperModel("tiny")
+    model = WhisperModel("base")
     pipeline = BatchedInferencePipeline(model=model)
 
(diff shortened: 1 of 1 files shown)


Purfview:
> can you please fix the failing tests?

Failing happens in `test_monotonic_timestamps()`, doesn't look like it's related to this PR.
Fails in the batched timestamps, maybe it reveals some bug in https://github.com/SYSTRAN/faster-whisper/pull/1297 if the test shouldn't fail.

Maybe tiny model can return crazy things AFAIK...

PR #1406, opened 2025-12-25
VAD Fix: explicit safe context, reduced RAM usage

Purfview:
VAD Fix: explicit safe context (no view mutation). 
Reduced SileroVADModel RAM usage by 9% [on 4h audio].

PR doesn't affect VAD speed nor probs output in current FW VAD functionality.
Fixes inconsistent outputs with multiple model calls [such functionality is currently not implemented].

### DIFF
--- a/faster_whisper/vad.py
+++ b/faster_whisper/vad.py
@@ -365,7 +365,9 @@ class SileroVADModel:
         batched_audio = np.empty((num_frames, frame_width), dtype=np.float32)
         batched_audio[:, context_size_samples:] = frames
         if num_frames > 1:
-            batched_audio[1:, :context_size_samples] = frames[:-1, -context_size_samples:]
+            batched_audio[1:, :context_size_samples] = frames[
+                :-1, -context_size_samples:
+            ]
         batched_audio[0, :context_size_samples] = 0.0
 
         encoder_batch_size = 10000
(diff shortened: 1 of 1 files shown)


MahmoudAshraf97:
Hi, I get that this may modify the array in place, but I don't understand how that affects multiple model calls, do you mean consecutive calls on the same audio array or multiple calls in parallel for the same or different audios?

Purfview:
>do you mean consecutive calls on the same audio array or multiple calls in parallel for the same or different audios?

As I remember it affected both consecutive and parallel calls with different size audios.

Not sure now, I think it affected same size audios too but only(?) the last prob was affected.

Purfview:
Analyzed probs from batched VAD [with 100% overlap] vs non-batched

Pre patch stats:
```
Total probs             : 224965
Total diff probs        : 112396
Above > 0.00001 diffs   : 641
Above > 0.0001 diffs    : 210
Above > 0.001 diffs     : 54
Above > 0.01 diffs      : 0
```

After the patch:
```
Total probs             : 224965
Total diff probs        : 97207
Above > 0.00001 diffs   : 0
Above > 0.0001 diffs    : 0
Above > 0.001 diffs     : 0
Above > 0.01 diffs      : 0 
```

Off-topic:
I'm leaning towards small hardcoded overlap, need to do more analysis, then I'll make a PR for batched VAD.

PR #1412, opened 2026-01-14
feat(audio): add soundfile fast path for WAV/FLAC decoding

eschmidbauer:
The existing PyAV-based decode_audio() has significant overhead:
- Creates AudioResampler for every file
- Calls gc.collect() after each decode
- Uses FFmpeg pipeline even for simple WAV files

This adds a fast path using soundfile (libsndfile) that:
- Reads WAV/FLAC files directly without FFmpeg overhead
- Falls back to PyAV for unsupported formats or file-like objects
- Provides 10-100x speedup for batch transcription workloads

Tested on 1M synthetic audio files where decode time dropped from
6-45 seconds per file to milliseconds.

Requires: pip install soundfile (optional, graceful fallback if missing)


MahmoudAshraf97:
how long were the files included in the 1M test?

PR #1425, opened 2026-03-05
Update README.md to add vod2verticalclips-converter-transcriber to the list

n7z2:


n7z2:
Wanted to add my personal project to the community list that is using the faster-whisper library. My tool is useful for content creators or anyone who quickly wants to create short form content i.e. YouTube shorts or Instagram reels from their videos. 

The tool provides an option to auto generate captions from the video which uses the faster-whisper library to achieve this.

Issue #1427, opened 2026-03-15
`BatchedInferencePipeline` silently overrides caller-provided timestamp options and diverges from `WhisperModel.transcribe()` when `without_timestamps=False`


ponpaku:
## Summary

`BatchedInferencePipeline.transcribe()` with `without_timestamps=False` is not behaviorally equivalent to `WhisperModel.transcribe()` under the same API-level inputs.

There are two parts that look reportable at upstream issue scope:

1. **Confirmed bug**: the batched timestamp path silently overrides caller-provided:
   - `condition_on_previous_text`
   - `max_initial_timestamp`
2. **Observed divergence**: even with the same model and user-provided options, the batched timestamp path can diverge substantially from the direct path on public fixtures, including:
   - fewer segments than direct
   - more segments than direct
   - worse CER than direct on some fixtures

I first noticed this as transcript-completeness degradation on non-public Japanese audio, but the issue below uses only public, reconstructable fixtures.

## Public repro source

I built public fixtures from Google FLEURS Japanese:

- dataset: `google/fleurs`
- config: `ja_jp`
- split: `test`

Primary fixture:

- sample ids: `1670`, `1937`, `1735`, `1716`, `1743`, `1852`
- inserted silence between clips: `800 ms`

Alternate fixtures:

- `alt1`: separate concatenated FLEURS Japanese fixture
- `alt2`: separate concatenated FLEURS Japanese fixture

Local files used to build and evaluate the fixtures:

- `public_repro/build_public_repro.py`
- `public_repro_manifest.json`
- `public_repro/fixtures/public_repro_manifest_alt1.json`
- `public_repro/fixtures/public_repro_manifest_alt2.json`
- `public_repro/run_public_repro.py`

## Environment

- faster-whisper version: `1.2.1`
- upstream base commit tested locally: `ed9a06c`
- device: `cpu`
- compute type: `int8`
- language: `ja`
- `vad_filter=True`
- `without_timestamps=False`

### Build the fixture

I used a local script to concatenate the public FLEURS clips into a single WAV fixture:

```bash
python public_repro/build_public_repro.py
```

### Compare direct vs batched

```python
from faster_whisper import BatchedInferencePipeline, WhisperModel

model = WhisperModel("small", device="cpu", compute_type="int8")
pipeline = BatchedInferencePipeline(model=model)

direct_segments, direct_info = model.transcribe(
    "public_repro_fixture.wav",
    language="ja",
    vad_filter=True,
[...]

PR #1428, opened 2026-03-16
Respect caller timestamp options in batched transcribe

ponpaku:
# Summary

This PR restores caller-provided `condition_on_previous_text` and `max_initial_timestamp` in `BatchedInferencePipeline.transcribe()`.

Related issue: #1427

Previously, the batched path hardcoded these values instead of using the caller request, making it behaviorally non-equivalent to `WhisperModel.transcribe()` for timestamped decoding.

## What this PR changes

- pass through `condition_on_previous_text` in batched transcription options
- pass through `max_initial_timestamp` in batched transcription options
- add regression tests that verify both values are preserved in batched transcription options

## What this PR does not try to solve

This PR is intentionally narrow.

It does **not** attempt to fix all direct/batched quality gaps such as:

- prompt/history update design
- fallback behavior differences
- VAD chunk packing strategy

## Validation

- added lightweight regression tests for option propagation
- verified the new tests pass locally with `pytest`
- confirmed on a public FLEURS Japanese fixture that upstream batched transcription reports the overridden values, while the patched branch preserves the caller values

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -359,14 +359,14 @@ class BatchedInferencePipeline:
             condition_on_previous_text: If True, the previous output of the model is provided
                 as a prompt for the next window; disabling may make the text inconsistent across
                 windows, but the model becomes less prone to getting stuck in a failure loop,
-                such as repetition looping or timestamps going out of sync. Set as False
+                such as repetition looping or timestamps going out of sync.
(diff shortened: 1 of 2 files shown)


MahmoudAshraf97:
This PR adds confusion, the arguments you mentioned are not used, so the true values are correctly represented in transcription options regardless of the user provided values, which, as I already nention, are not used

PR #1436, opened 2026-04-13
fix: pin ctranslate2<4.6.3 to maintain CUDA compatibility

### LINKED ISSUE #1413: Ctranslate import break retro compatibility in 4.6.3
Hi,

There is a contamination bug from ctranslate2 dependancy: 
https://github.com/OpenNMT/CTranslate2/issues/1978

They decided to exclude device under CUDA 12.8 in 4.6.3 patch version.

The requirements of faster whisper lib should then block upgrade of current version to 4.6.2 to maintain retro compatibility.

Regards,
Pierre

Anai-Guo:
## Problem

ctranslate2 4.6.3 dropped support for CUDA devices below 12.8 (OpenNMT/CTranslate2#1978). Since faster-whisper specifies `ctranslate2>=4.0,<5`, users on older CUDA versions get broken installations when pip resolves to 4.6.3+.

## Fix

Pin `ctranslate2>=4.0,<4.6.3` in `requirements.txt` to maintain backward compatibility.

Fixes #1413

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-ctranslate2>=4.0,<5
+ctranslate2>=4.0,<4.6.3
 huggingface_hub>=0.23
 tokenizers>=0.13,<1
 onnxruntime>=1.14,<2 


Anai-Guo:
Closing — the upstream pin in requirements.txt was already updated to `ctranslate2>=4.0,<5`, which addresses the same compatibility concern with a different upper bound. Apologies for the noise.

PR #1440, opened 2026-04-19
Add audio_backend parameter (pyav default, ffmpeg alternative)

andreluisrabelo:
## Summary

Adds a new `audio_backend` keyword argument to `WhisperModel.__init__` (and to the lower-level `decode_audio()` helper) that lets callers opt out of PyAV and decode audio via a subprocess call to the `ffmpeg` binary instead.

The default remains `"pyav"` — **no behavior change for existing users**.

## Motivation

PyAV ships precompiled wheels on PyPI. On Windows 11 with Smart App Control (SAC) enabled, CodeIntegrity blocks the PyAV binaries (`pyio.pyd`, `libwebpmux-*.dll`) as not meeting the "Enterprise signing level" requirement. `import av` then crashes with DLL load failures (Event ID 3077 in CodeIntegrity/Operational).

SAC cannot be disabled without reinstalling Windows, which makes `faster-whisper` effectively unusable on corporate / hardened desktops — even though a signed `ffmpeg.exe` (e.g. from `winget install Gyan.FFmpeg`) is readily available on those same machines.

The same escape hatch is useful in other sandboxed contexts (locked-down corporate images, minimal containers) where a system `ffmpeg` is present but arbitrary Python-shipped DLLs cannot be loaded.

Related: the workaround we currently run in production (patching `site-packages/faster_whisper/audio.py` after install) — this PR upstreams that logic behind a clean public flag so nobody needs to monkey-patch the wheel.

## Proposed API

```python
from faster_whisper import WhisperModel

# Default, unchanged behavior
model = WhisperModel("large-v3")

# Opt in to ffmpeg subprocess backend
model = WhisperModel("large-v3", audio_backend="ffmpeg")
```

The same parameter is also accepted on the lower-level helper for callers who use it directly:

```python
from faster_whisper import decode_audio

audio = decode_audio("speech.mp3", backend="ffmpeg")
```

Accepted values: `"pyav"` (default) or `"ffmpeg"`. Anything else raises `ValueError` eagerly — both in `decode_audio` and at `WhisperModel` construction time, before any weights are downloaded.

### ffmpeg binary resolution

When `audio_backend="ffmpeg"` is selected, the binary is resolved in this order:

1. `FFMPEG_EXE` environment variable (absolute path).
2. `shutil.which("ffmpeg")` on `PATH`.
3. Otherwise, raise `RuntimeError` with an install hint.

## Implementation notes

- `faster_whisper/audio.py` is refactored so that `import av` happens **lazily** inside the pyav code path. Users who pick `audio_backend="ffmpeg"` never trigger a PyAV import, which is exactly what SAC-locked environments need.
- The ffmpeg path runs:
  ```
  ffmpeg -nostdin -threads 0 -i <input> -f s16le -ac 1 -acodec pcm_s16le -ar <rate> -loglevel error -
  ```
  and converts the raw PCM stdout to `float32 / 32768.0` via numpy, matching the pyav pipeline byte-for-byte up to resampler rounding.
- File-like inputs are supported via `stdin` piping (mirroring the current pyav behavior).

## Trade-offs

- `audio_backend="ffmpeg"` requires a system `ffmpeg`. This is already a common dependency for many faster-whisper workflows, and it is the only viable path on SAC-protected Windows installs.
- Default stays `"pyav"` — no new runtime dependency unless explicitly opted into.

## Tests

New `tests/test_audio_backend.py` covers:

- Default `pyav` backend decodes the existing `jfk.flac` fixture to `float32` (smoke).
- `ffmpeg` backend produces a matching waveform (length drift < 20 ms, RMS delta < 1e-2) — skipped when `ffmpeg` is not on the runner's `PATH`.
- Invalid backend values are rejected by `decode_audio` and by `WhisperModel.__init__` (validation happens before weight download).
- Missing `ffmpeg` binary is reported as `RuntimeError` with an install hint — simulated via monkeypatched `shutil.which` so the test runs deterministically regardless of host state.

Local run on macOS (M-series, Python 3.9):

```
tests/test_audio_backend.py::test_supported_backends_listed PASSED
tests/test_audio_backend.py::test_decode_audio_default_backend_pyav PASSED
tests/test_audio_backend.py::test_decode_audio_ffmpeg_backend_matches_pyav PASSED
tests/test_audio_backend.py::test_decode_audio_invalid_backend_raises PASSED
tests/test_audio_backend.py::test_whisper_model_invalid_audio_backend_raises PASSED
tests/test_audio_backend.py::test_decode_audio_ffmpeg_backend_missing_binary_raises PASSED
6 passed in 15.53s
```

## README

Short snippet added next to the existing PyAV note, documenting when/why to pick `audio_backend="ffmpeg"` and how to point at a non-PATH binary via `FFMPEG_EXE`.

## Status

**Opened as DRAFT for upstream review.** I will not promote it to Ready until a maintainer has taken a look at the API shape. Happy to adjust — for example, if you'd prefer the flag on `transcribe()` instead of `__init__`, or a different default-resolution strategy for the ffmpeg binary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,14 @@ For reference, here's the time and memory usage that are required to transcribe 
 
 Unlike openai-whisper, FFmpeg does **not** need to be installed on the system. The audio is decoded with the Python library [PyAV](https://github.com/PyAV-Org/PyAV) which bundles the FFmpeg libraries in its package.
 
+If PyAV cannot be loaded in your environment (for example on Windows 11 with Smart App Control enabled, which blocks unsigned DLLs shipped by the PyAV wheel), you can opt into the alternative `ffmpeg` audio backend, which shells out to a system `ffmpeg` binary instead:
+
+```python
+model = WhisperModel("large-v3", audio_backend="ffmpeg")
+```
+
+This requires `ffmpeg` to be available on `PATH`, or the `FFMPEG_EXE` environment variable to point at its absolute path. The default remains `audio_backend="pyav"`, so nothing changes for existing users.
+
 ### GPU
 
 GPU execution requires the following NVIDIA libraries to be installed:
--- a/faster_whisper/audio.py
+++ b/faster_whisper/audio.py
@@ -1,25 +1,37 @@
-"""We use the PyAV library to decode the audio: https://github.com/PyAV-Org/PyAV
+"""Audio decoding with pluggable backends.
 
-The advantage of PyAV is that it bundles the FFmpeg libraries so there is no additional
-system dependencies. FFmpeg does not need to be installed on the system.
+By default, audio is decoded via the Python library `PyAV`
+(https://github.com/PyAV-Org/PyAV), which bundles FFmpeg so no system
+dependency is required.
 
-However, the API is quite low-level so we need to manipulate audio frames directly.
+An alternative `ffmpeg` backend is also available. It shells out to the
+`ffmpeg` binary via `subprocess`. This is useful on environments where the
+PyAV wheels cannot be loaded (e.g. Windows 11 Smart App Control blocking
+unsigned binaries, locked-down corporate environments, or sandboxes that
+forbid loading arbitrary DLLs) but where a signed/system `ffmpeg` binary is
+available on PATH. Select it via the `audio_backend='ffmpeg'` parameter on
+`WhisperModel` (or by calling `decode_audio(..., backend='ffmpeg')`).
 """
 
 import gc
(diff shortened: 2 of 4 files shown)


MahmoudAshraf97:
Hello and thanks for the contribution, I understand the motivation, but I don't think adding another backend is the solution, you should just move the `pyav` import to avoid triggering it, and load the audio however you want, `transcribe` method accepts mono 16khz numpy audio arrays

Issue #1320, opened 2025-07-11
Unable to install Faster Whisper


Denis-Kazakov:
OS: Linux Mint. Dedicated  conda environment

After installing Cublas and cudnn with `pip install nvidia-cublas-cu12 nvidia-cudnn-cu12==9.*`,  I ran:


export LD_LIBRARY_PATH=`python3 -c 'import os; import nvidia.cublas.lib; import nvidia.cudnn.lib; print(os.path.dirname(nvidia.cublas.lib.__file__) + ":" + os.path.dirname(nvidia.cudnn.lib.__file__))'`


which gave me the following error:
```
Traceback (most recent call last):                                                                                                                                                                                                            
  File "<string>", line 1, in <module>                                                                                                                                                                                                        
  File "<frozen posixpath>", line 152, in dirname                                                                                                                                                                                             
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

Generally, installation is very brittle. I was able to install Faster Whisper after installing transformers first (without explicitly installing cublas/cuddn), but I am unable to repeat the trick on the same computer a week later.

Denis-Kazakov:
That's a pity because otherwise it is a good tool with useful features, e.g. its built-in VAD helped me to solve a hallucination problem.
(Running it in a Docker container did the trick with CUDA and hopefully will be reproducible.)

gregtzar:
According to a search: In recent versions of nvidia-cublas-cu12 and nvidia-cudnn-cu12, NVIDIA changed how these packages are structured. The submodules are namespaces or do not explicitly declare a __file__ property anymore, returning None across the board

Anyways this updated command works:
```bash
export LD_LIBRARY_PATH=$(python3 -c 'import os, nvidia.cublas, nvidia.cudnn; print(os.path.dirname(nvidia.cublas.__path__[0]) + "/cublas/lib:" + os.path.dirname(nvidia.cudnn.__path__[0]) + "/cudnn/lib")')
```

I submitted a PR to fix this: https://github.com/SYSTRAN/faster-whisper/pull/1459

cc: @Denis-Kazakov

PR #1459, opened 2026-06-10
Fixes broken export LD_LIBRARY_PATH cmd in README.md

gregtzar:
The current cmd to export the `LD_LIBRARY_PATH` gave the following error on my machine:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "<frozen posixpath>", line 181, in dirname
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

This has been reported in issue: https://github.com/SYSTRAN/faster-whisper/issues/1320 and this update fixes the error.

Reason:  In recent versions of nvidia-cublas-cu12 and nvidia-cudnn-cu12, NVIDIA changed how these packages are structured. The submodules are namespaces or do not explicitly declare a __file__ property anymore, returning None across the board.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ On Linux these libraries can be installed with `pip`. Note that `LD_LIBRARY_PATH
 ```bash
 pip install nvidia-cublas-cu12 nvidia-cudnn-cu12==9.*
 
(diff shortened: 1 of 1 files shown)


MahmoudAshraf97:
Hi and thanks for the contribution, however, the correct place to fix this is in CTranslate 2, it should be able to accept the pypi packages without having to point it explicitly, then we can remove the wrong export completely

gregtzar:
@MahmoudAshraf97 I see, that would certainly be ideal. Hopefully in the meantime though we could at least fix the import paths that are currently required so it doesn't confuse anyone else.

Issue #1451, opened 2026-06-01
decode_audio crashes with "Frame does not match AudioFifo parameters" on files whose sample rate/format changes mid-stream


raivisdejus:
`decode_audio` fails on audio files whose stream changes sample rate (or sample format) partway through. `_group_frames` creates an `AudioFifo()` and writes every  decoded frame into it, but the FIFO locks onto the parameters of the first frame and rejects any later frame that differs:

```
  File "faster_whisper/audio.py", line 96, in _group_frames
      fifo.write(frame)
    File "av/audio/fifo.pyx", line 81, in av.audio.fifo.AudioFifo.write
  ValueError: Frame does not match AudioFifo parameters.
```

The AudioResampler later in the chain would happily normalize these frames, but the crash happens in _group_frames, before resampling.

Sample audio that produces the issue.
[whisper-french-mixed-samplerate.mp3](https://github.com/user-attachments/files/28475840/whisper-french-mixed-samplerate.mp3)

This issue has been reported in my project by user. The sample file above is a synthetic/crafted example, but apparently such issue can happen in the wild. 
Love faster whisper, excellent tool! 
Best regards and much energy in future development of faster whisper!

PR #1462, opened 2026-06-28
[Fix] decode_audio crashes on mid-stream sample rate/format changes (#1451)

yushuosun:
## Motivation

`decode_audio` raises `ValueError: Frame does not match AudioFifo parameters` on audio whose sample rate or sample format changes partway through the stream (#1451, which includes a repro file).

## Root cause

The decode pipeline groups frames *before* resampling:

```python
frames = _ignore_invalid_frames(frames)
frames = _group_frames(frames, 500000)        # AudioFifo locks to the first frame's params
frames = _resample_frames(frames, resampler)
```

`_group_frames` writes every decoded frame into a single `av.audio.fifo.AudioFifo()`, which fixes its parameters on the first frame and rejects any later frame with a different sample rate/format. So the crash happens in `_group_frames`, before the `AudioResampler` (which would normalize the frames) ever runs.

## Modifications

`faster_whisper/audio.py` (`decode_audio`): resample before grouping. `_resample_frames` normalizes every frame to `s16` / the target layout / `sampling_rate`, so `_group_frames` then only ever sees uniform frames. The `AudioResampler` is stateful across calls, so resampling frame-by-frame before batching yields the same samples as the previous order.

## Duplicate-check

- Track A — `gh api 'repos/SYSTRAN/faster-whisper/pulls?state=open --paginate'` grepped for `1451` → no open PR references the issue.
- Track B — issue #1451 (state OPEN) has 0 comments and no in-progress claim.

### DIFF
--- a/faster_whisper/audio.py
+++ b/faster_whisper/audio.py
@@ -46,8 +46,8 @@ def decode_audio(
     with av.open(input_file, mode="r", metadata_errors="ignore") as container:
         frames = container.decode(audio=0)
         frames = _ignore_invalid_frames(frames)
-        frames = _group_frames(frames, 500000)
         frames = _resample_frames(frames, resampler)
+        frames = _group_frames(frames, 500000)
 
         for frame in frames:
             array = frame.to_ndarray()


yushuosun:
Closing after prolonged repository inactivity and no maintainer review. The fix would also need the requested regression test and a performance check before it is ready to land; the branch remains available if development resumes.

Issue #1474, opened 2026-08-13
TypeError: __init__() should return None, not 'NoneType' in SpeechTimestampsMap under Python 3.14


keelung-yang:
<img width="1745" height="790" alt="Image" src="https://github.com/user-attachments/assets/82e2ac2d-5a13-400a-a640-8f77cd856994" />

Python version: 3.14.6, venv:

- faster-whisper version: 4.8.1, model: large
- ctranslate2-4.8.1
- nvidia_cublas_cu12-12.9.2.10
- nvidia_cudnn_cu12-9.24.0.43
- nvidia_cuda_nvrtc_cu12-12.9.86
- onnxruntime-1.28.0
- tokenizers-0.23.1

OS: CachyOS, all packages updated
```text
Operating System: CachyOS Linux 
KDE Plasma Version: 6.7.4
KDE Frameworks Version: 6.28.0
Qt Version: 6.11.1
Kernel Version: 7.1.8-1-cachyos (64-bit)
Graphics Platform: Wayland
Processors: 36 × Intel® Xeon® CPU E5-2695 v4 @ 2.10GHz
Memory: 32 GiB of RAM (31.3 GiB usable)
Graphics Processor: NVIDIA GeForce RTX 5060 Ti
Manufacturer: Dell Inc.
Product Name: Precision Tower 7810
```

keelung-yang:
Another `NoneType` issue:
```text
Exception in thread Exception in thread tqdm_monitortqdm_monitor:
:
Traceback (most recent call last):
Traceback (most recent call last):
  File "/usr/lib/python3.14/threading.py", line 1082, in _bootstrap_inner
  File "/usr/lib/python3.14/threading.py", line 1082, in _bootstrap_inner
        self._context.run(self.run)self._context.run(self.run)

  File "/usr/lib/python3.14/site-packages/tqdm/_monitor.py", line 78, in run
  File "/usr/lib/python3.14/site-packages/tqdm/_monitor.py", line 78, in run
        instances = self.get_instances()instances = self.get_instances()

  File "/usr/lib/python3.14/site-packages/tqdm/_monitor.py", line 58, in get_instances
  File "/usr/lib/python3.14/site-packages/tqdm/_monitor.py", line 58, in get_instances
    return [i for i in self.tqdm_cls._instances.copy()    
return [i for i in self.tqdm_cls._instances.copy()  File "/usr/lib/python3.14/_weakrefset.py", line 52, in copy

[...]

PR #1484, opened 2026-09-03
fix: disable tqdm's background monitor thread (mitigates #1474)

Bryandero98:
## Investigation

`SpeechTimestampsMap.__init__` (`faster_whisper/vad.py:283`) is plain Python with no return statement, so `TypeError: __init__() should return None, not 'NoneType'` cannot legitimately come from that class — this rules out a bug in that code itself.

The key clue is the second traceback attached to #1474: the *exact same* impossible error, independently raised from CPython's own stdlib (`_weakrefset.WeakSet.copy()` calling `self.__class__(self)`), inside `tqdm`'s `TMonitor` background thread. Two unrelated `__init__` calls hitting the identical, logically-impossible error is strong evidence this is a rare CPython 3.14 interpreter/threading race (most likely GC interacting with a background thread mid-instantiation), not something specific to faster-whisper's code.

I installed Python 3.14.7 (both the standard build and the free-threaded/no-GIL build) and stress-tested this directly: many threads concurrently instantiating `SpeechTimestampsMap`, `_weakrefset.WeakSet`, and a couple of unrelated plain classes, alongside threads forcing continuous `gc.collect()`, for 90s under the free-threaded build (where this kind of race is most likely to surface, since the GIL normally serializes access). I could not reproduce the error in either build. I don't take this as evidence the race doesn't exist — the original report is on Linux (CachyOS) and the failure is, by nature, a rare timing-dependent race that may depend on details I can't replicate here (OS threading behavior, the real C extensions in play - onnxruntime and ctranslate2 - versus my pure-Python stand-in for the class shape, or something else entirely). I could not find a matching upstream CPython issue either. So I want to be explicit: I have not confirmed the exact mechanism, only ruled out that it's a defect in `SpeechTimestampsMap`'s own code.

## What this PR does

`tqdm`'s monitor thread only exists to detect stalled progress bars (`monitor_interval`, documented in `tqdm/std.py`, "set to 0 to disable the thread") and serves no purpose in this library. Disabling it removes one background thread from the process for its entire lifetime, which directly avoids the failure mode demonstrated in that second traceback.

**This is a mitigation, not a confirmed root-cause fix.** Given I couldn't reproduce the underlying race even with the matching Python version, I can't verify this mitigation actually prevents #1474 from recurring - but removing an unnecessary background thread is a safe, real reduction of risk surface regardless, and has no user-visible behavior change (the progress bars themselves are untouched; only stalled-bar *detection* is disabled). Happy to close this if the maintainers would rather wait for a confirmed repro before merging a mitigation.

## Test plan

- [x] `pytest tests/` — 19/19 passing (Python 3.13, Windows)
- [x] `black .` / `isort .` / `flake8 .` — clean on the changed file
- [x] Stress-tested the underlying `__init__`/GC/threading race for 90s on Python 3.14.7, standard and free-threaded builds — no reproduction in either (see Investigation above)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/faster_whisper/transcribe.py
+++ b/faster_whisper/transcribe.py
@@ -27,6 +27,15 @@ from faster_whisper.vad import (
     get_speech_timestamps,
 )
 
+# tqdm's background monitor thread (used to detect stalled progress bars) has
+# been observed to trigger a rare CPython interpreter race under Python 3.14
+# (an unrelated `__init__` call elsewhere in the process fails with
+# "TypeError: __init__() should return None, not 'NoneType'", see
+# https://github.com/SYSTRAN/faster-whisper/issues/1474). It serves no
+# purpose here, so disable it outright rather than leave a background thread
+# running for the lifetime of the process.
+tqdm.monitor_interval = 0
+
 
 @dataclass
 class Word:


keelung-yang:
Disabling a feature to bypass an existed bug, it's not a good idea, I think.

Bryandero98:
Fair concern, and I want to be precise about what this PR actually does (and doesn't do), since "disabling a feature to bypass a bug" isn't quite the shape of it:

- The bug isn't in a *feature* of faster-whisper, tqdm, or even `SpeechTimestampsMap` - the traceback shows the exact same impossible error (`__init__()` returning non-`None`) independently raised from two unrelated places, including CPython's own stdlib (`_weakrefset.WeakSet.copy()`). That pattern points to a rare Python 3.14 interpreter/threading-GC race, not a defect any of us can patch in application code.
- What's disabled is tqdm's *monitor* thread specifically - a background watchdog that exists only to detect a stalled progress bar and force a refresh. The progress bars themselves are untouched: they render, update, and complete exactly as before. Nothing user-facing changes.
- I was explicit in the PR description that this is a mitigation (remove a background thread that serves no purpose here, reducing exposure to the race), not a confirmed fix - I stress-tested the suspected race on Python 3.14.7 (standard and free-threaded builds) and couldn't reproduce it, so I can't prove this prevents #1474 from recurring.

If you'd rather this wait until someone can actually reproduce and root-cause the race (or if a maintainer disagrees with the approach), I'm fine closing this - just didn't want "disabling a feature" to be read as removing something you'd notice using the library. Happy to explore other angles if you have one in mind.

